PR #342, opened 2015-01-27
Updating TfIdfIndex class in tfidf.py to work around hash collisions

Fideln8:
added self._id_to_i dictionary to hold a reverse pairing of key/values used in self._i_to_id.
updated _hash32 to check for the hash value in the dictionary keys before adding and updating the hash if necessary.
updated the index function to assign values to both dictionaries.
updated unindex to pull the value for i from the _id_to_i dictionary to ensure it is getting a record_id that exists.

Let me know if you have any questions.


fgregg:
The fix to hash32 looks good but Your code has diverged in a extraneous ways from master. Please clean that up.

PR #343, opened 2015-01-27
Update tfidf.py

Fideln8:
Updated class TfIdfIndex to handle hash collisions that were possible with the _hash32 function. 
added self._id_to_i dictionary to hold a reverse pairing of key/values used in self._i_to_id.
updated _hash32 to check for the hash value in the dictionary keys before adding and updating the hash if necessary.
updated the index function to assign values to both dictionaries.
updated unindex to pull the value for i from the _id_to_i dictionary to ensure it is getting a record_id that exists.

I reforked and added these changes. Let me know if you have any questions.

### DIFF
--- a/dedupe/tfidf.py
+++ b/dedupe/tfidf.py
@@ -21,7 +21,7 @@ class TfIdfIndex(object) :
     def _hash32(self, x) :
         i = hash(x)
         key = int(math.copysign(i % (2**31), i))
-        while key in self._i_to_id and self._i_to_id[key] <> x:
+        while key in self._i_to_id and self._i_to_id[key] != x:
             key += 1
         return key
         
--- a/dedupe/tfidf.py
+++ b/dedupe/tfidf.py
@@ -13,17 +13,24 @@ class TfIdfIndex(object) :
         self._index = CanopyIndex(stop_words)
  
         self._i_to_id = {}
+        
+        self._id_to_i = {}
+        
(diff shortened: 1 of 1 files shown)


dedupe/tfidf.py
```diff
@@ -13,17 +13,24 @@ def __init__(self, field, stop_words=[]) :
         self._index = CanopyIndex(stop_words)
  
         self._i_to_id = {}
+        
+        self._id_to_i = {}
```
fgregg: This looks duplicated.

dedupe/tfidf.py
```diff
@@ -13,17 +13,24 @@ def __init__(self, field, stop_words=[]) :
...
+        
         self._parseTerms = self._index.lexicon.parseTerms
 
     def _hash32(self, x) :
         i = hash(x)
-        return int(math.copysign(i % (2**31), i))
+        key = int(math.copysign(i % (2**31), i))
+        while key in self._i_to_id and self._i_to_id[key] <> x:
```
fgregg: do `!=` not `<>`

dedupe/tfidf.py
fgregg: hmm.. if we are going to keep this mapping around, then it would be much simper to just get rid of _hash32 and do something like

`self._id_to_i = ids = collections.defaultdict(itertools.count(-2**31).next)`

PR #344, opened 2015-01-27
Update tfidf.py

Fideln8:
Update class TfIdfIndex to handle hash collisions that were possible with the _hash32 function.


fgregg:
oops needs to be 2**31 not 2**32, can you make the change?

fgregg:
Thanks!

PR #352, opened 2015-02-10
start on metric trees

fgregg:
Basically wired up. Going to need to make indexing and searching *much* faster before we can merge.


fgregg:
I'm thinking [aspell](http://aspell.net/)

fgregg:
fine-night is just too slow and takes too much memory.

Issue #285, opened 2014-07-22
Implement connectivity constrained clustering.


fgregg:
http://scikit-learn.org/stable/modules/clustering.html#adding-connectivity-constraints

https://github.com/scikit-learn/scikit-learn/blob/e6003bc892485756215ac4966b18720bc489e5c6/sklearn/cluster/hierarchical.py#L255

Hierarhical clustering [does not handle missing data well](https://github.com/datamade/dedupe/issues/195#issuecomment-49762359), and has a number of performance issues: #280, #281, #279.

fgregg:
@cathydeng, create a branch and swamp in connectivity constrained clustering from scikit. If it's good, we'll consider re-implementing in without depending upon scipy or scikit.

fgregg:
We might want to compare this to other segmentation approaches here: http://scikit-image.org/docs/0.10.x/auto_examples/plot_segmentations.html?highlight=cluster

cathydeng:
updates:

made a fix to sklearn so that agglomerative clustering works w/ a precomputed distance matrix & connectivity constraints: https://github.com/scikit-learn/scikit-learn/commit/fc51c062ce70e3e7c0083cf70ed8fc215b5535b9

b/c you need to specify the # of clusters as a parameter for agglomerative clustering, if we use it, we'll have to calculate clusters for various values of n_clusters, & select a value of n_clusters that makes the most sense (perhaps by using the threshold parameter in the dedupe cluster method?)

I'll continue working on implementing agglomerative clustering w/ connectivity constraints, & evaluating performance

fgregg:
Working the @cathydeng, I came to realization that this graph segmentation/community detection paradigms do *not* help us address the missing data problem that originally motivated us to look for something other than hierarchical agglomerative cluster.

Here's the problem in brief. Because of the vagaries of blocking, we could have made the following two pairwise comparisons.

```
(A,B), 0.7
(B,C), 0.3
```
But did not compare ```(A,C)```

In order to use hierarchical agglomerative clustering we need a full similarity matrix.
```
   A   B   C
A   1 0.7   ?
B 0.7   1 0.3
C   ? 0.3   1 
``` 
What do we put in the ``(A,C)`` cells? 

Right now we put zeros. But, that's not really right, the pairwise similarity between `A` and `C` is almost certainly not 0, but we have to put something in there to proceed with clustering.

[...]

PR #360, opened 2015-02-14
New clustering

fgregg:


fgregg:
These paradigms won't help with the missing data problem, see #285

PR #383, opened 2015-05-03
Make model training reproducible

shahin:
Reproducible models are useful in production and for debugging. But right now, trained models aren't reproducible given the same training data and data model because cross-validation folds are random. More training data or simpler models might make training more predictable but not deterministic.

```gridSearch``` currently has an unimplemented ```randomize``` parameter. This commit (a) replaces it with ```random_seed``` which defaults to None (current behavior), and (b) adds a random_seed attribute to ```ActiveMatching``` that's passed to ```gridSearch``` on training.

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -544,7 +544,8 @@ class ActiveMatching(Matching) :
     def __init__(self, 
                  variable_definition, 
                  data_sample = None,
-                 num_cores = None) :
+                 num_cores = None,
+                 random_seed = None) :
         """
         Initialize from a data model and data sample.
 
@@ -606,6 +607,8 @@ class ActiveMatching(Matching) :
             self.num_cores = multiprocessing.cpu_count()
         else :
             self.num_cores = num_cores
+
+        self.random_seed = random_seed
 
         training_dtype = [('label', 'S8'), 
                           ('distances', 'f4', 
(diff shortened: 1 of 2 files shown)


fgregg:
Can't the same thing be achieved by setting the random seed in before calling the dedupe train method?

shahin:
Yeah, calling ```numpy.random.seed()``` before the dedupe train method would work. That sets global state for everything that uses numpy.random in the same interpreter, though, which isn't ideal for systems that import dedupe too. Are you concerned about the new parameter?

fgregg:
I don't have a problem with adding a parameter to cross validate, but
adding a new parameter to the Dedupe.__init__ seems kind of heavy.

shahin:
I'm not excited about another init param either. I've thought about this more. Two options:

1. No change to master. Setting ```numpy.random.seed``` globally across every loaded module works just fine. Seems risky but maybe okay for now.
2. Double down on the new ```self.random_seed``` in Dedupe by changing ```randomPairs``` to take a ```random_seed``` param too. We get reproducibility, Dedupe instances each control all inputs to the training algorithm (including random numbers), and we don't affect other modules.

Anything I'm missing? Thanks for the review.

fgregg:
Sorry @shahin, let's just stick with setting numpy.random.seed for now.

PR #401, opened 2015-07-22
Added check for empty dictionary of records in RecordLink.sample

nmiranda:
Giving an empty dictionary as argument to 'RecordLink.sample' raises a cryptic 'ValueError' originating from the use of 'random.randrange' function, with zero as an argument, in 'sampling.py' module. This fix inserts a check for empty dictionary arguments in 'RecordLink.sample' function, and raises an appropriate exception message accordingly.


fgregg:
Thanks!

PR #402, opened 2015-07-23
Changed wrong default value '1.5' for argument 'threshold', in functi…

nmiranda:
Changed wrong default value '1.5' for argument 'threshold', in function 'RecordLinkMatching.match', to '0.5'.


fgregg:
Thanks!

PR #416, opened 2015-10-01
Cleanup process pool after initializing ActiveLearning

evz:
On machines with many processing cores, it's apparently necessary to terminate the process pool after using it in this case since all of the processes that this ``__init__`` method uses are still hanging around even if you delete the parent object.

### DIFF
--- a/dedupe/training.py
+++ b/dedupe/training.py
@@ -53,6 +53,8 @@ class ActiveLearning(object) :
             pool.map(fieldDistance, 
                      chunker(candidates, 100),
                      2))
+        
+        pool.terminate()
 
         self.seen_indices = set()
 


evz:
This is going to seem super weird but I actually had the same error that appveyor got when I was trying to setup a development version of dedupe in my local virtualenv. I eventually had to basically nuke everything and then install all of the requirements in setup.py using pip and then run setup.py develop

evz:
Here's the traceback (in case that appveyor thing goes away) https://gist.github.com/evz/efbc636358814b096bb8

fgregg:
Okay, I'll merge this an open up an issue to figure out why pool is not being garbage collected. When pool is garbage collected, then the terminate method should be automatically called. https://docs.python.org/3.4/library/multiprocessing.html#multiprocessing.pool.Pool.terminate

PR #426, opened 2015-11-24
Decouple details of classifier from data model

fgregg:
I'd like it easier to drop in a different classification module, but there is a tight coupling between the data model and classifier that makes that hard. There's not real reason that the data model should no about the classification weights.

This pull requests decouples the classifier and the datamodel. 

- dedupe also now expects the classifer object to mimic the interface of sklearn's classifiers, making it easy to drop in any sklearn classifiers
- fieldDistances is now a method of DataModel, instead of in core
- predicateGenerator is now a method of DataModel
- checkRecords is now a method of DataModel
- I've slimmed down the public attributes of DataModel
- cross validation has been moved to the classifer and out of the dedupe library


fgregg:
@evz there's a bunch of things that are going to be relevant for you here.

1. changed the interface for rlr to be a class, that uses `.fit` to train, stores it's own weights, and provides the `predict_proba` to score new examples. 
2. the datamodel object no longer stores field weights. the classifier stores its weights, and we now save the classifier object in the settings file. This means that we shouldn't show the weights on the settings page. 
3. the _logWeights method has been removed
4. the interface for datamodel, to get the field names, has changed https://github.com/datamade/dedupe/blob/learner_decouple/dedupe/datamodel.py#L24-L33

PR #435, opened 2015-12-30
Updated os-x install doc

lucaswiser:
This worked for me to get multiprocessing up and running on os-x. I think the main thing was also setting the [atlas] header to point to openblas.


fgregg:
Thanks!

PR #450, opened 2016-03-11
In consoleLabel(), present fields in the order they are defined.

wleftwich:


fgregg:
Really nice idea. I'm actually making some other changes to the console labeler right now, I'm going to close this PR, but incorporate this idea.

wleftwich:
If you are working on the labeler, maybe this will come in handy too:

def format_label_pair(pair, fieldnames, spacer='   '):
    lengths = {}
    for fieldname in fieldnames:
        lengths[fieldname] = max(len(str(y).strip()) for y in
(x[fieldname] for x in pair))
    L = []
    for item in pair:
        M = [str(item[fieldname]).strip().ljust(lengths[fieldname])
for fieldname in fieldnames]
        L.append(spacer.join(M))
    return '\n'.join(L)

Makes output like this:


stephen   bishop   wayne schwartz constructio    581443 grey rd 25
chesley   on   n0g 1l0

steven    bishop   wayne schwartz construction   rr 4
chesley   on   n0g 1l0


Do these records refer to the same thing?
(y)es / (n)o / (u)nsure / (f)inished

PR #459, opened 2016-04-07
Populate sample_records attributes from data_sample when it is supplied to Dedupe or RecordLink constructor.

wleftwich:
See issue https://github.com/datamade/dedupe/issues/458

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -849,7 +849,7 @@ class ActiveMatching(Matching) :
 
     def _loadSampledRecords(self, data_sample):
         """Override to load blocking data from data_sample."""
-        pass
+
 
 
 class StaticDedupe(DedupeMatching, StaticMatching) :
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -849,7 +849,7 @@ class ActiveMatching(Matching) :
 
     def _loadSampledRecords(self, data_sample):
         """Override to load blocking data from data_sample."""
-        self.sampled_records = None
+        pass
 
 
 class StaticDedupe(DedupeMatching, StaticMatching) :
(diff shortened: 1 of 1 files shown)


fgregg:
Thanks @wleftwich this looks like a great start. We also need to think about how to set the `original_length` attribute of `Sample` 

I'm not sure where the right place is, maybe an argument to train. 

I think I'll probably accept this pull request and think about the `original_length` problem a little longer.

dedupe/api.py
```diff
@@ -845,6 +848,13 @@ def _loadSample(self, data_sample) :
                                                      self.data_model,
                                                      self.num_cores)
 
+    def _load_sampled_records(self, data_sample, sample_size=900):
+        if data_sample:
+            recs = itertools.chain.from_iterable(data_sample)
```
fgregg: This will work for Dedupe class but not for RecordLink classes, where we will need to set `self.sampled_records_1` and `self.sampled_records_2`

wleftwich: Oh right. I will try moving the sampled_records code down into `Dedupe` and `RecordLink`.

wleftwich:
I have moved the code for creating `sampled_records` from `data_sample` down into the `Dedupe` and `RecordLink` constructors -- so the title of this PR is no longer accurate.

dedupe/api.py
```diff
@@ -861,7 +859,17 @@ class Dedupe(DedupeMatching, ActiveMatching) :
     """
     canopies = True
 
-    def sample(self, data, sample_size=15000, 
+    def __init__(self, *args, **kwargs):
+        super(Dedupe, self).__init__(*args, **kwargs)
+        data_sample = kwargs.get('data_sample')
+        if data_sample:
+            recs = itertools.chain.from_iterable(data_sample)
```
fgregg: How about having most of this code happen in the ActiveMatching __init__ and then having methods in Dedupe and RecordLink like `_sampledRecords`

wleftwich: Sure, OK. I did enjoy typing `super(Dedupe` though. Can probably commit changes by tomorrow morning.

fgregg:
Looks good!

PR #461, opened 2016-04-12
Alternate persistence

fgregg:


fgregg:
Pretty close, we just need to handle gazetteer case

PR #479, opened 2016-08-17
load data into defaultdict

jbrissier:
I had an issue if i'am using the StaticGazetteer. When i save the index to a file and load it later, i was not able to index new data. (Get a KeyError). The pickle library loads the do_to_id as python dict instant of a defaultdict. I use the factory from the TfIdfIndex, but i am not 100% sure if this is the right place to this.

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -548,7 +548,7 @@ class StaticMatching(Matching):
             for predicate in full_predicate:
                 if hasattr(predicate, "index") and predicate.index is None:
                     predicate.index = predicate.initIndex()
-                    doc_to_id_max_id = numpy.amax([v for k, v in doc_to_ids[predicate].items()])
+                    doc_to_id_max_id = max(doc_to_ids[predicate].values())
                     predicate.index._doc_to_id = defaultdict(itertools.count(doc_to_id_max_id).next, doc_to_ids[predicate])
                     if hasattr(predicate, "canopy"):
                         predicate.canopy = canopies[predicate]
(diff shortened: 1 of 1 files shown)


dedupe/api.py
```diff
@@ -548,7 +548,8 @@ def _loadIndices(self, settings_file):
             for predicate in full_predicate:
                 if hasattr(predicate, "index") and predicate.index is None:
                     predicate.index = predicate.initIndex()
-                    predicate.index._doc_to_id = doc_to_ids[predicate]
+                    doc_to_id_max_id = numpy.amax([v for k, v in doc_to_ids[predicate].items()])
```
fgregg: this can just be ```max(doc_to_ids[predicate].values())``

dedupe/api.py
```diff
@@ -548,7 +548,8 @@ def _loadIndices(self, settings_file):
...
+                    predicate.index._doc_to_id = defaultdict(itertools.count(doc_to_id_max_id).next, doc_to_ids[predicate])
                     if hasattr(predicate, "canopy"):
```
fgregg: this needs to be 

```
defaultdict(itertools.count(doc_to_id_max_id+1).next, doc_to_ids[predicate])
```
to start counting **after** the current max id.

We also need to handle both python 2 and 3 like here: 
https://github.com/datamade/dedupe/blob/aa2d50af1a26be3f6ae03a262c1b750484a1f473/dedupe/tfidf.py#L14-L17

fgregg:
Good catch!

jbrissier:
I will change the stuff

fgregg:
Thanks!

fgregg:
Whoops, had some change blindness. https://github.com/datamade/dedupe/commit/db4c628a472836f8b58beb56ad677b2942e87063

PR #497, opened 2016-11-10
Fix consoleLabel example in docs

toolness:
The current `consoleLabel` example in the docs appears to be broken:

```py
dedupe = Dedupe(variables)
dedupe.sample(data)
dedupe.consoleLabel(dedupe)
```

This code suggests that `consoleLabel` is an instance method of a `Dedupe` object, but in fact it appears to be a function of the `dedupe` module, not an instance method.

This PR fixes the example by renaming the `dedupe` variable to `deduper` so that it doesn't shadow the `dedupe` module.

### DIFF
--- a/docs/API-documentation.rst
+++ b/docs/API-documentation.rst
@@ -263,7 +263,7 @@ Convenience Functions
 
    .. code:: python
 
-      > deduper = Dedupe(variables)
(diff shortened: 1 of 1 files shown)


fgregg:
Thanks

PR #500, opened 2016-11-23
Update requirements.txt

gautam1858:
Please add this in the requirement its useful


fgregg:
This shouldn't be necessary, since it's already listed as a dependency in setup.py.

Issue #478, opened 2016-07-22
EOF error in python 2.7 (not in python 3.5)


titipata:
I got an error when running `consoleLabel` on Jupyter notebook (my Jupyter version is `4.1.0`) using Python 2.7 as follows.

```python
dedupe.consoleLabel(deduper)
```

Here is an error:

```
EOFError                                  Traceback (most recent call last)
<ipython-input-41-5f16b23568e5> in <module>()
----> 1 dedupe.consoleLabel(deduper)

~/anaconda2/lib/python2.7/site-packages/dedupe/convenience.pyc in consoleLabel(deduper)
     55             while not valid_response:
     56                 print('(y)es / (n)o / (u)nsure / (f)inished', file=sys.stderr)
---> 57                 label = input()
     58                 if label in ['y', 'n', 'u', 'f']:
     59                     valid_response = True

EOFError: EOF when reading a line
```

I'm wondering if there is an error for others who use Python 2.7 also. Or, the error is raised because of Jupyter notebook. I see no error when using Python 3.5.

fgregg:
This seems like a Jupyter problem.

titipata:
I think so too. We can close the issue if it's not related. I'm just wondering if there are other people who has the same problem running dedupe on Jupyter notebook.

nateyoder:
I'm getting this issue as well. Did you ever find a work around @titipata ?  Thanks!

titipata:
Hello @nateyoder! it has been quite a while on this issue. I don't remember how I solved an issue actually.  I'm not sure if changing [convenience.py#L56](https://github.com/datamade/dedupe/blob/372716493b99fc7f89f9e142c0c5a694420a0484/dedupe/convenience.py#L56) from `input` to `raw_input` will help solve the problem or not. In this case, you might have to download `dedupe`, fix L56, and use `setup.py` in order to install the library.

nateyoder:
All good.  I ended up getting it to work by basically doing exactly what you suggested by replacing [line 4](https://github.com/datamade/dedupe/blob/372716493b99fc7f89f9e142c0c5a694420a0484/dedupe/convenience.py#L4) with:

```python
try:
    input = raw_input
except NameError:
    pass
```

Thanks!

PR #509, opened 2017-01-10
Use raw_input to help with input errors in jupyter notebook

nateyoder:
Fallback to using input in python3.

Created based off my issues and #478

### DIFF
--- a/dedupe/convenience.py
+++ b/dedupe/convenience.py
@@ -1,7 +1,10 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 from __future__ import print_function
-from builtins import input
+try:
+    input = raw_input
+except NameError:
+    pass
 
 import collections
 import itertools


fgregg:
Sorry, I don't want to take on the responsibility for support jupyter

nateyoder:
Fair enough but I really think this is just best practice when dealing with Python2 and Python3: http://stackoverflow.com/questions/21731043/use-of-input-raw-input-in-python-2-and-3

That said I'm happy to maintain my own fork :)  Thanks for a great package.

fgregg:
Hmm, didn't know this was a python 2/3 issue.

 I would accept this http://stackoverflow.com/a/38691215/98080

fgregg:
Nevermind, that's already what I'm doing, haha. Good luck.

nateyoder:
Yeah I believe that is what you had but at least for me that doesn't actually work.

fgregg:
Strangely, python-future should do exactly what your PR did. https://github.com/PythonCharmers/python-future/blob/master/src/future/builtins/misc.py#L57

You could open up an issue there to figure out why it didn't.

PR #518, opened 2017-02-09
add lacking packages to requirements.txt

yatszhash:
add lacking dependencies to requirements.txt for developers.


fgregg:
Those dependencies are already in setup.py, we don't need to duplicate them in requirements.txt.

PR #523, opened 2017-02-21
Added 'previous' option to manual labelling (`consoleLabel`)

Leobouloc:
When using the command line interface to label data, I sometimes made mistakes requiring me to start labeling all over again (or to manually delete text from the `training.json` file). As this could get somewhat frustrating, I added a "(p)revious" command that allows users to go back to labelling previous labels if one are several mistakes are made.

Interface looks as follows:

```
### Previous is possible
(y)es / (n)o / (u)nsure / (f)inished / (p)revious

### Previous is not possible 
(y)es / (n)o / (u)nsure / (f)inished

### Previous is not possible but p was given as input
(p)revious is not a valid answer (no record in memory)
(y)es / (n)o / (u)nsure / (f)inished
```

This solution uses a fixed length buffer to store examples. These examples are marked (using `deduper.markPairs(examples)`) only once the buffer is full, allowing the user to go back as many steps as the buffer length. By default, I set the buffer length to 1 in the code (`buffer_len = 1`, `consoleLabel`) not being sure what was the appropriate way to add parameters to this function (through deduper, or new variable). The left over data in the buffer is marked at the end of training

Also, I assumed `deduper.uncertainPairs()` would always return a single pair, so I added an assertion to check that and removed a loop on `uncertain_pairs` (instead: `record_pair = uncertain_pairs[0]`).

I hope this can be useful in any way. This is my first pull request on a public project; please let me know if I am doing something wrong :)

### DIFF
--- a/dedupe/convenience.py
+++ b/dedupe/convenience.py
@@ -30,7 +30,7 @@ def consoleLabel(deduper): # pragma: no cover
                     in deduper.data_model.primary_fields)
 
 
-    buffer_len = 0 # Max number of previous operations
+    buffer_len = 1 # Max number of previous operations
     record_label_buffer = []
     uncertain_pairs = []
 
--- a/dedupe/convenience.py
+++ b/dedupe/convenience.py
@@ -30,21 +30,21 @@ def consoleLabel(deduper): # pragma: no cover
                     in deduper.data_model.primary_fields)
 
     buffer_len = 1 # Max number of previous operations
-    record_label_buffer = []
+    examples_buffer = []
     uncertain_pairs = []
 
     while not finished :       
-        if len(record_label_buffer) and use_previous:
-            uncertain_pairs.append(record_label_buffer[0][0]) # append previous record_pair
-            record_label_buffer = record_label_buffer[1:] # remove previous pair from buffer
+        if len(examples_buffer) and use_previous:
+            record_pair = examples_buffer.pop(0)[0]
+        else:
+            if not uncertain_pairs:
+                uncertain_pairs = deduper.uncertainPairs()
+            record_pair = uncertain_pairs.pop()
+             
         use_previous = False
         
-        if not uncertain_pairs:
-            uncertain_pairs = deduper.uncertainPairs()
-        record_pair = uncertain_pairs[-1]
-                
-        n_match_in_buffer = sum(x[1]=='match' for x in record_label_buffer) 
-        n_distinct_in_buffer = sum(x[1]=='distinct' for x in record_label_buffer)
+        n_match_in_buffer = sum(label=='match' for _, label in examples_buffer) 
+        n_distinct_in_buffer = sum(label=='distinct' for _, label in examples_buffer)
                                                                                         
         n_match, n_distinct = (len(deduper.training_pairs['match']) + n_match_in_buffer,
                         len(deduper.training_pairs['distinct']) + n_distinct_in_buffer)
@@ -62,49 +62,46 @@ def consoleLabel(deduper): # pragma: no cover
         print('Do these records refer to the same thing?', file=sys.stderr)
         valid_response = False
         while not valid_response:
-            if len(record_label_buffer):
+            if len(examples_buffer):
                 print('(y)es / (n)o / (u)nsure / (f)inished / (p)revious', file=sys.stderr)
-                label = input()
-                if label in ['y', 'n', 'u', 'f', 'p']:
+                user_input = input()
+                if user_input in ['y', 'n', 'u', 'f', 'p']:
                     valid_response = True
             else:
                 print('(y)es / (n)o / (u)nsure / (f)inished', file=sys.stderr)
-                label = input()
-                if label in ['y', 'n', 'u', 'f']:
+                user_input = input()
+                if user_input in ['y', 'n', 'u', 'f']:
                     valid_response = True
-                if label == 'p':
-                    print('No record in memory: cannot use (p)revious', file=sys.stderr)
 
-        if label == 'y' :
-            record_label_buffer.insert(0, (record_pair, 'match'))
-            uncertain_pairs.pop() # Remove current pair form uncertain list
-        elif label == 'n' :
-            record_label_buffer.insert(0, (record_pair, 'distinct'))
-            uncertain_pairs.pop() 
-        elif label == 'f':
-            print('Finished labeling', file=sys.stderr)
-            finished = True
-        elif label == 'u':
-            record_label_buffer.insert(0, (record_pair, 'uncertain'))
-            uncertain_pairs.pop() 
-        elif (label == 'p') and len(record_label_buffer):
-            use_previous = True
+        if user_input == 'y' :
+            examples_buffer.insert(0, (record_pair, 'match'))
+        elif user_input == 'n' :
+            examples_buffer.insert(0, (record_pair, 'distinct'))
+        elif user_input == 'u':
+            examples_buffer.insert(0, (record_pair, 'uncertain'))
         else:
-            print('Nonvalid response', file=sys.stderr)
-            raise
+            uncertain_pairs.append(record_pair)
+            if user_input == 'f':
+                print('Finished labeling', file=sys.stderr)
+                finished = True
+            elif (user_input == 'p') and len(examples_buffer):
+                use_previous = True
+            else:
+                print('Nonvalid response', file=sys.stderr)
+                raise
         
-        if len(record_label_buffer) > buffer_len:
-            (record_pair, true_label) = record_label_buffer.pop()
-            if true_label in ['distinct', 'match']:
+        if len(examples_buffer) > buffer_len:
+            (record_pair, label) = examples_buffer.pop()
+            if label in ['distinct', 'match']:
                 examples = {'distinct' : [], 'match' : []}
-                examples[true_label].append(record_pair)
+                examples[label].append(record_pair)
                 deduper.markPairs(examples)
        
-    for (record_pair, true_label) in record_label_buffer:
-        if true_label in ['distinct', 'match']:
+    for (record_pair, label) in examples_buffer:
+        if label in ['distinct', 'match']:
             examples = {'distinct' : [], 'match' : []}
-            examples[true_label].append(record_pair)
-    deduper.markPairs(examples)
+            examples[label].append(record_pair)
+            deduper.markPairs(examples)
 # 
 
 def trainingDataLink(data_1, data_2, common_key, training_size=50000) : # pragma: nocover
--- a/dedupe/convenience.py
+++ b/dedupe/convenience.py
@@ -34,21 +34,20 @@ def consoleLabel(deduper): # pragma: no cover
     uncertain_pairs = []
 
     while not finished :       
-        if len(examples_buffer) and use_previous:
-            record_pair = examples_buffer.pop(0)[0]
+        if examples_buffer and use_previous:
+            record_pair, _ = examples_buffer.pop(0)
+            use_previous = False
         else:
             if not uncertain_pairs:
                 uncertain_pairs = deduper.uncertainPairs()
             record_pair = uncertain_pairs.pop()
-             
-        use_previous = False
-        
+                     
         n_match_in_buffer = sum(label=='match' for _, label in examples_buffer) 
         n_distinct_in_buffer = sum(label=='distinct' for _, label in examples_buffer)
                                                                                         
         n_match, n_distinct = (len(deduper.training_pairs['match']) + n_match_in_buffer,
                         len(deduper.training_pairs['distinct']) + n_distinct_in_buffer)
-      
+        
         label = ''
 
         for pair in record_pair:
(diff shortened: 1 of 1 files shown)


fgregg:
By moving markPairs out of the loop, we are no longer doing active learning.

Leobouloc:
Agreed. In this case, though, `markPairs` is not really out of the loop, it is just delayed by the buffer (examples are marked once they leave the buffer: [this line in the code](https://github.com/datamade/dedupe/pull/523/commits/ad325bc9a382f87d986305235802d32c816cff43#diff-b114ead7ac17865996f06abddf2de266R105)). I thought this could be acceptable if the buffer is not too big: it's max length is currently set to 1, meaning we are just missing out on the last label for active learning.

dedupe/convenience.py
```diff
@@ -24,55 +24,94 @@ def consoleLabel(deduper): # pragma: no cover
...
+
+    while not finished :     
+            
+        if has_previous and use_previous:
+            record_pair = record_pair_buffer[0]
+            record_pair_buffer = record_pair_buffer[1:]
+            examples_buffer = examples_buffer[1:]
+            use_previous = False
+        else:
+            uncertain_pairs = deduper.uncertainPairs()
```
fgregg: I would prefer the previous pattern that did not make assumptions about the length of uncerttain_pairs.

dedupe/convenience.py
```diff
@@ -24,55 +24,94 @@ def consoleLabel(deduper): # pragma: no cover
...
 
-    while not finished :
-        n_match, n_distinct = (len(deduper.training_pairs['match']),
-                               len(deduper.training_pairs['distinct']))
-
-        uncertain_pairs = deduper.uncertainPairs()
-
+    buffer_len = 1 # Max number of previous operations
+    examples_buffer = []
+    record_pair_buffer = []
```
fgregg: Might be cleaner to make `example_buffer` a list of tuples of labels and the record_pair. You can turn these into the format that markPairs expects at the last moment. 

If you did this you wouldn't need two buffers.

dedupe/convenience.py
```diff
@@ -24,55 +24,94 @@ def consoleLabel(deduper): # pragma: no cover
...
+            record_pair = record_pair_buffer[0]
+            record_pair_buffer = record_pair_buffer[1:]
+            examples_buffer = examples_buffer[1:]
+            use_previous = False
+        else:
+            uncertain_pairs = deduper.uncertainPairs()
+            assert len(uncertain_pairs) == 1 
+            record_pair = uncertain_pairs[0]
+       
+        n_match_in_buffer = sum(len(x['match']) for x in examples_buffer)
```
fgregg: if you follow my suggestion above then this would be something like. 

```n_match_in_buffer = sum(label == 'distinct' for label, pair in example_buffer)```

File: dedupe/convenience.py [hunk 4]
```diff
@@ -24,55 +24,94 @@ def consoleLabel(deduper): # pragma: no cover
...
+                line = "%s : %s" % (field, pair[field])
+                print(line, file=sys.stderr)
+            print(file=sys.stderr) 
+
+        print("{0}/10 positive, {1}/10 negative".format(n_match, n_distinct),
+                file=sys.stderr)
+        print('Do these records refer to the same thing?', file=sys.stderr)
+        valid_response = False
+        while not valid_response:
+            if has_previous:
```
fgregg: make the prompt and valid responses variables and you can reduce the code duplication here.

dedupe/convenience.py
```diff
@@ -24,55 +24,94 @@ def consoleLabel(deduper): # pragma: no cover
...
-            elif label == 'f':
-                print('Finished labeling', file=sys.stderr)
-                finished = True
-            elif label != 'u':
-                print('Nonvalid response', file=sys.stderr)
-                raise
-
-        if labeled :
-            deduper.markPairs(examples)
+                if label == 'p':
```
fgregg: Why is this special message necessary?

dedupe/convenience.py
```diff
@@ -24,55 +24,94 @@ def consoleLabel(deduper): # pragma: no cover
...
+            print('Finished labeling', file=sys.stderr)
+            finished = True
+        elif (label == 'p') and has_previous:
+            use_previous = True
+        elif label != 'u':
+            print('Nonvalid response', file=sys.stderr)
+            raise
         
+        if labeled:
+            examples_buffer = [examples] + examples_buffer
```
fgregg: this should be examples_buffer.append(example)

dedupe/convenience.py
```diff
@@ -24,55 +24,94 @@ def consoleLabel(deduper): # pragma: no cover
...
+            record_pair_buffer = [record_pair] + record_pair_buffer
+            assert len(examples_buffer) == len(record_pair_buffer)
```
fgregg: you should be able to do away with two buffers.

dedupe/convenience.py
```diff
@@ -24,55 +24,94 @@ def consoleLabel(deduper): # pragma: no cover
...
+
+            if len(examples_buffer) > buffer_len:
+                examples = examples_buffer.pop()
+                record_pair_buffer.pop()
```
fgregg: This is where you should construct the example in the format that markPairs exists.

dedupe/convenience.py
```diff
@@ -24,55 +24,94 @@ def consoleLabel(deduper): # pragma: no cover
     '''
 
     finished = False
+    has_previous = False
```
fgregg: You don't need this variable. Where you are using this you can just do

```if len(example_buffer)```

**Comment by Leobouloc** [comment]
Thank you for the feedback. I'll make the changes when I get a chance.

File: dedupe/convenience.py (hunk 5 above)
Comment (Leobouloc) [inline dedupe/convenience.py]: I put this for the user that would have seen the "previous" option once and who would not understand why the option would no longer be available. It could be removed though...

File: dedupe/convenience.py (hunk 6 above)
Comment (Leobouloc) [inline dedupe/convenience.py]: I replaced this by `buffer.insert(0, value)` in the following pull request. I can't use append here as I want to prepend the example (unless you were suggesting I invert the order of the elements in the buffer...).

File: dedupe/convenience.py (hunk 4 above)
Comment (Leobouloc) [inline dedupe/convenience.py]: I am not sure what code duplication we are trying to avoid...

File: dedupe/convenience.py (hunk 9 above)
Comment (Leobouloc) [inline dedupe/convenience.py]: changed in the next commit

File: dedupe/convenience.py (hunk 1 above)
Comment (Leobouloc) [inline dedupe/convenience.py]: changed in the next commit (also fixes -i think- the potential problem of not being able to `(f)inish` if in the middle of the `for record pair in uncertain_pairs` loop)

File: dedupe/convenience.py (hunk 6 above)
Comment (fgregg) [inline dedupe/convenience.py]: why does the order in the buffer matter?

File: dedupe/convenience.py (hunk 5 above)
Comment (fgregg) [inline dedupe/convenience.py]: i think w don't need it?

File: dedupe/convenience.py (hunk 6 above)
Comment (Leobouloc) [inline dedupe/convenience.py]: Currently, I insert the elements that were just labeled on the left, and use `markPairs` for the element on the right of the buffer when it is full (the buffer can potentially hold several values). Elements in the buffer (`list`) are sorted by order of labeling. Does this answer your question?

File: dedupe/convenience.py [hunk 10]
```diff
@@ -24,55 +24,88 @@ def consoleLabel(deduper): # pragma: no cover
...
 
-    while not finished :
-        n_match, n_distinct = (len(deduper.training_pairs['match']),
-                               len(deduper.training_pairs['distinct']))
-
-        uncertain_pairs = deduper.uncertainPairs()
-
-        examples = {'distinct' : [], 'match' : []}
-
-        for record_pair in uncertain_pairs:
```
fgregg: This is the pattern I want to keep.

dedupe/convenience.py
```diff
@@ -24,55 +24,88 @@ def consoleLabel(deduper): # pragma: no cover
...
-
-            print("{0}/10 positive, {1}/10 negative".format(n_match, n_distinct),
-                  file=sys.stderr)
-            print('Do these records refer to the same thing?', file=sys.stderr)
-            valid_response = False
-            while not valid_response:
+    while not finished :       
+        if len(record_label_buffer) and use_previous:
+            uncertain_pairs.append(record_label_buffer[0][0]) # append previous record_pair
+            record_label_buffer = record_label_buffer[1:] # remove previous pair from buffer
```
fgregg: These two lines could just be

```
uncertain_pairs.append(record_label_buffer.pop()[0])
```

dedupe/convenience.py
```diff
@@ -24,55 +24,88 @@ def consoleLabel(deduper): # pragma: no cover
     '''
 
     finished = False
+    use_previous = False
     fields = unique(field.field
                     for field
                     in deduper.data_model.primary_fields)
 
+    buffer_len = 1 # Max number of previous operations
+    record_label_buffer = []
```
fgregg: this seems like a poor name since it's not just a buffer of labels. What was wrong with `examples`

dedupe/convenience.py
```diff
@@ -24,55 +24,88 @@ def consoleLabel(deduper): # pragma: no cover
...
-            while not valid_response:
+    while not finished :       
+        if len(record_label_buffer) and use_previous:
+            uncertain_pairs.append(record_label_buffer[0][0]) # append previous record_pair
+            record_label_buffer = record_label_buffer[1:] # remove previous pair from buffer
+        use_previous = False
+        
+        if not uncertain_pairs:
+            uncertain_pairs = deduper.uncertainPairs()
+        record_pair = uncertain_pairs[-1]
```
fgregg: Why not do

```record_pair = uncertaing_pairs.pop()```

here? Seems like it would simplify things later.

File: dedupe/convenience.py [hunk 14]
```diff
@@ -24,55 +24,88 @@ def consoleLabel(deduper): # pragma: no cover
...
+                
+        n_match_in_buffer = sum(x[1]=='match' for x in record_label_buffer)
```
fgregg: I think using tuple unpacking would make this clearer.

```
sum(label==='match' for _, label in record_label_buffer)
```

dedupe/convenience.py
```diff
@@ -24,55 +24,88 @@ def consoleLabel(deduper): # pragma: no cover
...
-                raise
-
-        if labeled :
-            deduper.markPairs(examples)
+                if label == 'p':
+                    print('No record in memory: cannot use (p)revious', file=sys.stderr)
+
+        if label == 'y' :
+            record_label_buffer.insert(0, (record_pair, 'match'))
+            uncertain_pairs.pop() # Remove current pair form uncertain list
```
fgregg: If you pop above, when you get record_pair you don't need all these pops down here.

dedupe/convenience.py
```diff
@@ -24,55 +24,88 @@ def consoleLabel(deduper): # pragma: no cover
...
+            uncertain_pairs.pop() 
+        elif (label == 'p') and len(record_label_buffer):
+            use_previous = True
+        else:
+            print('Nonvalid response', file=sys.stderr)
+            raise
         
+        if len(record_label_buffer) > buffer_len:
+            (record_pair, true_label) = record_label_buffer.pop()
+            if true_label in ['distinct', 'match']:
```
fgregg: how about just `label`? `true_label` makes me think that there are false labels somewhere?

dedupe/convenience.py
```diff
@@ -24,55 +24,88 @@ def consoleLabel(deduper): # pragma: no cover
...
+                examples = {'distinct' : [], 'match' : []}
+                examples[true_label].append(record_pair)
+                deduper.markPairs(examples)
+       
+    for (record_pair, true_label) in record_label_buffer:
+        if true_label in ['distinct', 'match']:
+            examples = {'distinct' : [], 'match' : []}
+            examples[true_label].append(record_pair)
+    deduper.markPairs(examples)
```
fgregg: it seems possible that examples will never be defined? what about that case?

dedupe/convenience.py
Leobouloc: I meant it as records and labels but I'll change it back to examples

dedupe/convenience.py
Leobouloc: yes. I will put `deduper.markPairs(examples)` back in the `if` statement

dedupe/convenience.py
Leobouloc: I wanted to have distinct names for the user input (`y`, `n`, etc.) and the actual label (`match`,`distinct`)... I can change `label` to `user_input` and `true_label` to `label`. Would that work ?

dedupe/convenience.py
Leobouloc: I chose this to avoid having to re-append `record_pair` to `uncertain_pairs` if we don't get a valid user input but changing it would indeed remove the three `pop` statements.

dedupe/convenience.py
fgregg: Got it. good idea!

dedupe/convenience.py
fgregg: how about something like

```python
if record_label_buffer and use_previous:
    record_pair =  record_label_buffer.pop[0]
else:
    if not uncertain_pairs:
        uncertain_pairs = deduper.uncertainPairs()
    record_pair = uncertain_pairs.pop()    
```

dedupe/convenience.py
Leobouloc: Replaced this by `uncertain_pairs.append(record_label_buffer.pop(0)[0])`

fgregg: if you are appending at front of record_label_buffer, you need to pop from the end, not the front.

Leobouloc: These lines correspond to the situation where `previous` was chosen so we want to add the last labeled element to the list of `uncertain_pairs`. Since I append at front of `record_label_buffer`, in this case I want to pop(0) to retrieve the latest element.

fgregg: got it!

fgregg:
Thank you, @Leobouloc. This will be helpful.

PR #535, opened 2017-04-01
shelve backing for blocking

fgregg:


fgregg:
@amarder, this is the approach that I think might be worth investigating.

amarder:
Awesome!

PR #548, opened 2017-04-20
fix dedupeio/dedupe#547: handle empty filtered_pairs

yuanxu-li:
fix dedupeio/dedupe#547
to resolve the out-of-index error

### DIFF
--- a/dedupe/core.py
+++ b/dedupe/core.py
@@ -77,7 +77,7 @@ class ScoreRecords(object) :
 
             try :
                 filtered_pairs = self.fieldDistance(record_pairs)
-                if filtered_pairs is not None :
+                if filtered_pairs is not None and filtered_pairs.size != 0 :
                     score_queue.put(filtered_pairs)
             except Exception as e :
                 score_queue.put(e)
@@ -121,6 +121,7 @@ def mergeScores(score_queue, result_queue, stop_signals) :
 
     seen_signals = 0
     start = 0
+    end = 0
     while seen_signals < stop_signals  :
         score_chunk = score_queue.get()
         if isinstance(score_chunk, Exception) :


dedupe/core.py
```diff
@@ -121,6 +121,7 @@ def mergeScores(score_queue, result_queue, stop_signals) :
 
     seen_signals = 0
     start = 0
+    end = 0
```
fgregg: how would it be possible for end to not be defined in the following loop?

dedupe/core.py
```diff
@@ -77,7 +77,7 @@ def __call__(self, records_queue, score_queue) :
 
             try :
                 filtered_pairs = self.fieldDistance(record_pairs)
-                if filtered_pairs is not None :
+                if filtered_pairs is not None and filtered_pairs.size != 0 :
```
fgregg: I think I would prefer this to be handled by `self.fieldDistance` 

do a check of of `mask.any()` and return None if mask.any() is not true.

fgregg:
Thanks for this @yuanxu-li. I went a slightly differerent way 3114ce5072bb37d9775b88761b7d72cbb9af0233

dedupe/core.py
yuanxu-li: This is actually possible.
`score_queue` has only `None` because of a high threshold, and `score_chunk = score_queue.get()` gives out only a `None`.

In this case, the line that defines `end` (https://github.com/dedupeio/dedupe/blob/master/dedupe/core.py#L139) is never reached.

yuanxu-li: @fgregg Could you fix this as well?

fgregg: Okay, thanks for pointing out that.

PR #552, opened 2017-04-25
Dense blocks

fgregg:
@mbauman, this is an alternative way for making dense comparisons.  I'm liking this approach better.

Basically, the idea is that after blocking, we combine overlapping blocks and then compare all records within these merged blocks. 

So, if records 1, 2, and 3 are blocked together, and 3, 4, 5 are blocked together, we would combine these blocks into a block of (1, 2, 3, 4, 5).

I'm sure there's a way to do this sql, so if you want try to figure that out, that would be great.
Again, I've only implemented this for the in-memory case.

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -115,7 +115,6 @@ class Matching(object):
                                        self.classifier,
                                        self.num_cores,
                                        threshold=0)
-
         logger.debug("matching done, begin clustering")
 
         clusters = list(self._cluster(matches, threshold, *args, **kwargs))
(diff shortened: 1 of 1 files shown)


fgregg:
@mbauman here's a place to get started: 

```sql
with keys as (
  with a as (
    select *,count(1) over (partition by donor_id) c
    from plural_block
  )
  select block_id,'intersect'::text cc
  from a where c > 1
)
select array_agg(distinct donor_id)
from plural_block
left outer join keys on keys.block_id = plural_block.block_id
group by case when cc='intersect' then cc else plural_block.block_id end;
```

from https://stackoverflow.com/questions/43619332/unions-of-intersecting-sets?noredirect=1#comment74290358_43619332

fgregg:
@mbauman here's a start http://stackoverflow.com/a/21719104/98080

fgregg:
@mbauman Here's a start:

https://stackoverflow.com/questions/43619332/unions-of-intersecting-sets

fgregg:
By merging blocks that have any amount of intersection, we can create super blocks that are much too large to process.

Let's look at only merging blocks if there's already some threshold degree of overlap.

PR #554, opened 2017-04-26
Don't call _cluster on empty list

rlvoyer:
`scoreDuplicates` sometimes returns a `numpy.mmep` and other times returns an empty list. This
results in a type error in `connectedCommponents` which expects `edgelist` to be
dict-like. (Perhaps `scoreDuplicates` should return None rather than an empty list?)

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -118,7 +118,7 @@ class Matching(object):
 
         logger.debug("matching done, begin clustering")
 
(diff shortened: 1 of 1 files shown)


fgregg:
hi @rlvoyer thanks for this suggested fix. I ended up fixing this by making sure that the return value of scoreDuplicates was consistent.

PR #566, opened 2017-05-19
fix broken links in docs after name change from datamade to dedupeio

stevemartingale:
I noticed that the links under `Using dedupe` on the documentation page were broken so i fixed those and a few others


fgregg:
Thanks!

PR #570, opened 2017-06-02
Add dependency + docs for DateTime variable

jeancochrane:
This PR adds a dependency for `dedupe-variable-datetime` to `setup.py` and updates the docs, in order to bring the `DateTime` field type into dedupe core. 

(Also fixes a tiny little typo in the dedupe docs.)


docs/Variable-definition.rst
```diff
@@ -229,6 +229,35 @@ the value is 0 or negative, then an exception will be raised.
 
     {'field' : 'cost', 'type': 'Price'}
 
+DateTime
+^^^^^^^^
```
fgregg: Give indication of what type of values date time can be.

PR #580, opened 2017-06-28
try tuples instead of lists

fgregg:


fgregg:
@tendres, can you try this code, and can you also join me on the dedupe irc channel? https://webchat.freenode.net/?channels=dedupe

fgregg:
figure_1

This is the memory usage I'm getting during scoreDuplicates. notice that it jumps around, but on average does not really grow.

fgregg:
figure_tendres_1

Here's a run from tendres.

fgregg:
I'm also changing from SimpleQueue to a Queue of max size of 2 in the PR.

This changes the memory usage patten in a nice way. Here's the memory profile with SimpleQueue on a machine where multiprocessing is disabled.

simplequeue

And here it is with a size limited Queue.

limited_queue_size

Total memory usage doesn't really change in this case, but we are not creating this huge queue at the very beginning of the process.

PR #583, opened 2017-06-29
try to always use multiprocessing queues

fgregg:
right now, when we don't do multiprocessing we use the threading queue https://docs.python.org/3.4/library/queue.html#module-queue

Under the hood this is just a deque, which is probably holding on memory . Using the multiprocessing queues will both reduce complexity and hopefully help address memory issues on non-multiprocessing machines.

### DIFF
--- a/dedupe/backport.py
+++ b/dedupe/backport.py
@@ -26,16 +26,15 @@ elif platform.system() == 'Windows' :
     warnings.warn("Dedupe does not currently support multiprocessing on Windows")
     MULTIPROCESSING = False
 
+from multiprocessing import Queue
+if sys.version < '3':
+    from multiprocessing.queues import SimpleQueue
+else :
+    from multiprocessing import SimpleQueue
+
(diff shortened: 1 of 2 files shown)


fgregg:
I was puzzled about why my pattern of memory use during scoreDuplicates was different than @tendres. 

My memory use is pretty flat. (I killed this job before it got to clustering)

figure_1

While his memory use slowly increases (time 0 to time 350)
figure_1-1

It thought that it had to do with the fact that I am using multiprocessing and @tendres is not. In particular, if multiprocessing is enabled, the Queue is backed by a operating system Pipe process that use memory out of this current process. On the other hand, if multiprocessing is disabled, dedupe would use a Queue that was backed by a python `deque` object that lived in memory.

I thought that it was possible that this deque was holding onto memory, and that explained the difference. That does not appear to be the case as the memory profile that @tendres sent me as before. Multiprocessing may still be implicated, but it is nothing to do with the Queues.

PR #645, opened 2018-03-29
Fixed int overflow bug in randomPairsMatch 

vaneseltine:
Fixed int overflow bug in randomPairsMatch when using large datasets.

Added switch to numpy default (float_) when n_records_A*n_records_B > 2147483647

### DIFF
--- a/dedupe/core.py
+++ b/dedupe/core.py
@@ -71,6 +71,8 @@ def randomPairsMatch(n_records_A, n_records_B, sample_size):
 
     if sample_size >= n:
         random_pairs = numpy.arange(n)
+    elif n > numpy.iinfo(int).max:
+        random_pairs = numpy.array(random.sample(range(n), sample_size))
     else:
         random_pairs = numpy.array(random.sample(range(n), sample_size),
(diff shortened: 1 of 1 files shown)


dedupe/core.py
```diff
@@ -71,6 +71,8 @@ def randomPairsMatch(n_records_A, n_records_B, sample_size):
 
     if sample_size >= n:
         random_pairs = numpy.arange(n)
+    elif n > numpy.iinfo(int).max:
+        random_pairs = numpy.array(random.sample(range(n), sample_size))
```
fgregg: `i, j` need to be ints when this function returns. Will they be. 

If numpy.unravel_index does the right thing with floats, then we could simply this conditional by never doing a `dtype=int`

fgregg:
Hi @vaneseltine, do you have time to finish up this PR?

vaneseltine:
Sorry, getting a test together for a proper pull request turns out to be a
little more complicated than I anticipated, mostly due to my inexperience.
I'm afraid I don't immediately have time to learn right now.

PR #663, opened 2018-05-18
Added pool.terminate() in  core.py to fix thread leak I was experiencing.

Rehket:
The thread pool was not being cleared when score gazette was finished running. So if multiple gazetteer matches are called in a process, the threads were not being killed, and continued to accumulate.

I am using python 3.6.5 on Windows 10 with WSL running Ubuntu 16.04.

The code I was using is included here: https://github.com/Rehket/gazetteer-it


fgregg:
Thanks for the PR. I addressed this problem in another commit.

Rehket:
Great, Thank you!

PR #670, opened 2018-05-25
multimatching working!

fgregg:
## Background
In ["Dedoop: Efficient Deduplication with Hadoop"](https://dbs.uni-leipzig.de/file/Dedoop.pdf), Kolb, @andreas-thor, and Rahm propose a scheme to avoid expensive, redundant record comparisons. Their scheme can be reframed as follows:

When you block a record, attach all block_keys to the record.

```
record_id,keys
A,{1,3}
B,{1,3}
C,{2,3}
```

Then for each block, keep track of the block_keys for every record in the block that are less than or equal to the focal block key 

```
block_key, record_id, smaller_or_equal_block_keys
1,A,{1}
1,B,{1}
2,C,{2}
3,A,{1,3}
3,B,{1,3}
3,C,{2,3}
```

Finally, when it's time to compare records in a block, only compare them if the intersection of associated `smaller_block_keys` is equal to 1

So when we encounter (A,B) in block 1, we will compare the records.

But when we encounter (A,B) in block 3, we won't compare since the intersection of the associated blocks is `{1,3}` and has length of 2

But we will compare (B,C) in block 3, because `len({1,3} & {2,3)} == 1`

## Extension to multiple matching
Framed this way, it's natural to extend Kolb's scheme to only match if the intersection of block_keys is at least N instead of just 1.

For example, we may only want to compare a record pair if they share 4 words in common or 7 trigrams.

In order to achieve this, the block_keys need to include an identifier for the type of blocking rule and the number of matches required before matching.

Say block_key is a tuple of (blocking_rule_id, required_matches, block_predicate_value). Then the following algorithm will suffice:

```python
for block in blocks:
    for record_pair in block:
         record_1, record_2 = record_pair
         block_coverage = record_1.smaller_or_equal_blocks & record_2.smaller_or_equal_blocks
         current_block_key = max(block_coverage)
         current_blocking_rule_id = current_block_key[0]
         if should_compare(block_coverage, current_blocking_rule):
            score = expensive_compare(record_1, record_2)

def should_compare(block_coverage, current_blocking_rule_id):                  
     for grouped_cover in group_by_blocking_rule_id(block_coverage):
         blocking_rule_id, required_matches, _ = grouped_cover[0] # get info from arbitrary element of goruped_cover
         if blocking_rule_id != current_blocking_rule_id:
             if len(grouped_cover) >= required_matches:
                 return False #don't compare the records because they will be compared in another block with a smaller block key
         else:
             if len(grouped_cover) == required_matches:
                 return True # Now's the time compare
             else:
                 return False # Don't compare because either we haven't seen enough matches or seen too many 
```

## Jaccard Index
Prompted by a communication with Erhard Rahm, I investigated one avenue for extending this only comparing when a minimum threshold of the Jaccard index is met.

Other options
- attaching additional data to the record pairs such as cardinality of the blocks
- perhaps some of the preprocessing techniques
- using a special data structure to for filtering possible candidate records by cardinality (maybe an interval tree)

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -257,8 +257,10 @@ class DedupeMatching(Matching):
         self._checkBlock(block)
 
         combinations = itertools.combinations
+        product = itertools.product
 
-        pairs = (combinations(sorted(block), 2) for block in blocks)
+        pairs = (product((block_key,), combinations(sorted(block), 2))
+                 for block_key, block in blocks)
 
         return pairs
 
@@ -298,10 +300,11 @@ class DedupeMatching(Matching):
                                   if k < block_key}
                 processed_block.append((record_id, data_d[record_id], smaller_blocks))
 
-            yield processed_block
+            yield block_key, processed_block
 
     def _checkBlock(self, block):
         if block:
+            block_key, block = block
             try:
                 id, record, smaller_ids = block[0]
             except (ValueError, KeyError):
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -342,6 +342,89 @@ class RecordLinkMatching(Matching):
         self._cluster = clustering.greedyMatching
         self._linkage_type = "RecordLink"
 
+    def thresholdBlocks(self, blocks, recall_weight=1.5):  # pragma: nocover
+        """
+        Returns the threshold that maximizes the expected F score, a
+        weighted average of precision and recall for a sample of
+        blocked data.
+
+        Arguments:
+
+        blocks -- Sequence of tuples of records, where each tuple is a
+                  set of records covered by a blocking predicate
+
+        recall_weight -- Sets the tradeoff between precision and
+                         recall. I.e. if you care twice as much about
+                         recall as you do precision, set recall_weight
+                         to 2.
+
+        """
+        candidate_records = itertools.chain.from_iterable(self._blockedPairs(blocks))
+
+        probability = core.scoreRecordLink(candidate_records,
+                                           self.data_model,
+                                           self.classifier,
+                                           self.num_cores)['score']
+
+        probability = probability.copy()
+        probability.sort()
+        probability = probability[::-1]
+
+        expected_dupes = numpy.cumsum(probability)
+
+        recall = expected_dupes / expected_dupes[-1]
+        precision = expected_dupes / numpy.arange(1, len(expected_dupes) + 1)
+
+        score = recall * precision / (recall + recall_weight ** 2 * precision)
+
+        i = numpy.argmax(score)
+
+        logger.info('Maximum expected recall and precision')
+        logger.info('recall: %2.3f', recall[i])
+        logger.info('precision: %2.3f', precision[i])
+        logger.info('With threshold: %2.3f', probability[i])
+
+        return probability[i]
+
+    def matchBlocks(self, blocks, threshold=.5, *args, **kwargs):
+        """
+        Partitions blocked data and generates a sequence of clusters,
+        where each cluster is a tuple of record ids
+
+        Keyword arguments:
+
+        blocks -- Sequence of tuples of records, where each tuple is a
+                  set of records covered by a blocking predicate
+
+        threshold -- Number between 0 and 1 (default is .5). We will
+                      only consider as duplicates record pairs as
+                      duplicates if their estimated duplicate
+                      likelihood is greater than the threshold.
+
+                      Lowering the number will increase recall,
+                      raising it will increase precision
+
+        """
+        candidate_records = itertools.chain.from_iterable(self._blockedPairs(blocks))
+
+        matches = core.scoreRecordLink(candidate_records,
+                                       self.data_model,
+                                       self.classifier,
+                                       self.num_cores,
+                                       threshold=0)
+
+        logger.debug("matching done, begin clustering")
+
+        for cluster in self._cluster(matches, threshold, *args, **kwargs):
+            yield cluster
+
+        try:
+            match_file = matches.filename
+            del matches
+            os.remove(match_file)
+        except AttributeError:
+            pass
+        
     def match(self, data_1, data_2, threshold=0.5, generator=False):  # pragma: no cover
         """
         Identifies pairs of records that refer to the same entity, returns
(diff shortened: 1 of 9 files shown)


fgregg:
todo

- [ ] improve block generator code
- [ ] we now have scoreDuplicates, scoreRecordLink (which is almost exactly scoreDuplicates, and scoreGazetteer, rationalize this (probably in separate PR)
- [ ] think about making this transition as smooth as possible
- [ ] possibilities for grouping by in SQL? 
- [ ] dedupe database examples

fgregg:
attn @vm-wylbur and @resteorts (this is proving to be a very nice improvement to blocking)

fgregg:
At the moment of block creation, we know how many block_keys a overlap predicate will produce.

If we know that, and we know a minimum threshold we will required for the jaccard index, can
we make a kind of compound index that where the second element is the range of cardinalities that *could* produce a minimum threshold if at least matches were made?

fgregg:
### Attempt 1, Hashing approach
Start by observing that we can make a blocking rule that allows blocks together nearby integers, by creating block_keys that range around the focal integer. Say we have records the contain the age of persons, and we want to compare people who have recorded ages within 2 years of one another. We can do this by taking each age and making a block keys for each number in the range 

`[focal age - 2, focal age + 2]`

So, we have the following three records

```
record_id,age
1,25
2,30
3,32
```

this would generate blocking like 

```
block_key,record_id
23,1
24,1
25,1
26,1
27,1
28,2
29,2
30,2
31,3
32,3
30,3,
31,3
32,3
33,3
34,3
```
records 2 and 3 have overlapping block keys so we will compare those records. We won't compare 1 and 2 or 1 and 3 because there are no overlapping block keys.

Next, observe that for a given set **A** with cardinality N then the smallest possible set **B** that could have a particular Jaccard Index value would have cardinality  ```ceiling(N * (Jaccard Index value)```, and the largest set could have cardinality ```floor(N / (Jaccard Index value)```.

For example, say we have a set **A** of 4 elements and we have a desired Jaccard Index value of 0.5. The smallest set that could do that would have 2 elements and the largest set could be have 8 elements.

At the time of blocking, we know how many block_keys are created by a block predicate. We can use these relationships above to create compound blocks to filter out record pairs that could not have the desired minimum Jaccard Index value.

So say we have a tokenizing predicate that generates the following block_keys

```
{'the', 'lazy', 'fox', 'jumped'}
```

We'd like to only compare only compare records where the Jaccard Index of the tokens is 0.5 or greater.

We can then create the following block_keys:

```
{'the:2', 'lazy:2', 'fox:2', 'jumped:2',
 'the:3', 'lazy:3', 'fox:3', 'jumped:3',
 'the:4', 'lazy:4', 'fox:4', 'jumped:4',
 'the:5', 'lazy:5', 'fox:5', 'jumped:5',
 'the:6', 'lazy:6', 'fox:6', 'jumped:6',
 'the:7', 'lazy:7', 'fox:7', 'jumped:7',
 'the:8', 'lazy:8', 'fox:8', 'jumped:8'}
```

However, using the multiple matching scheme described above, we can do better than this and vary the number of required matches before comparison. Again, assuming we want a Jaccard Index of 0.5, the required matches are expressed by the following relation

```
minimum_required_matches = floor((Jaccard Index)/(1 + Jaccard Index)(|A| + |B|))
```

Here again are the block keys along with associated minimum matches to get a Jaccard index of 0.5

```
{('the:2', 2), ('lazy:2', 2), ('fox:2', 2), ('jumped:2', 2),
 ('the:3', 3), ('lazy:3', 3), ('fox:3', 3), ('jumped:3', 3),
 ('the:4', 3), ('lazy:4', 3), ('fox:4', 3), ('jumped:4', 3),
 ('the:5', 3), ('lazy:5', 3), ('fox:5', 3), ('jumped:5', 3),
 ('the:6', 4), ('lazy:6', 4), ('fox:6', 4), ('jumped:6', 4),
 ('the:7', 4), ('lazy:7', 4), ('fox:7', 4), ('jumped:7', 4)
 ('the:8', 4), ('lazy:8', 4), ('fox:8', 4), ('jumped:8', 4)}
```

This approach works, but as is probably evident it produced an enormous multiplication of block keys and blocks. As such, it does not seems currently practical for cases where there will be more than three or four original block keys. I have a rough implementation here: https://github.com/dedupeio/dedupe/tree/jaccard_hashing

fgregg:
Upon further reflection, we can simplify the block_keys and required matches to just

```
{('the', 2), ('lazy', 2), ('fox', 2), ('jumped', 2),
 ('the', 3), ('lazy', 3), ('fox', 3), ('jumped', 3),
 ('the', 4), ('lazy', 4), ('fox', 4), ('jumped', 4)}
```

This is still a significant multiplication, but probably tractable.

fgregg:
No we can't:

Counterexample: the lazy dog hopped. JI index of 2/6 <= 0.5

fgregg:
it might be worth doing this though as it pretty closely approximates the jaccard?

fgregg:
The maximum error for this approach is 
```
(1 - j) * j / (2 - j) 
```
that is if our target JI threshold is then we may accept sets that have a JI index of 

```
j - (1 - j) * j / (2 - j)
```

```
target, possible acceptance
0.1, 0.05
0.2, 0.11
0.3, 0.18
0.4, 0.25
0.5, 0.33
0.6, 0.43
0.7, 0.54
0.8, 0.66
0.9, 0.82
```

fgregg:
So.. that's one pathway. An alternative path, is to take this as an opportunity for a locally sensitive hashing scheme. I looked into this years ago, but it did not fit well with existing hashing paradigm.

A LSH approach could be very nice as it removes the relationship between the number of elements and the number of block_keys.

fgregg:
We can further reduce the **blocks** to the following
```
{('lazy', 2), ('fox', 2), ('jumped', 2),
 ('fox', 3), ('jumped', 3),
 ('jumped', 4)}
```

So, let n be the cardinality of a set of block_keys the number of blocks we need to make will be

```0.5 * (n/2 + 1) * (n/2 + 2)```

this is still quadratic.   

```
n,num_blocks
2,3
4,6
6,10
8,15
10,21 
20,66
30,136
```

fgregg:
This ended up not really be worth the creation of a lot more blocks for most cases.

PR #722, opened 2019-01-14
Fix OverflowError in randomPairsMatch()

az0:
Closes https://github.com/dedupeio/dedupe/issues/677

PR #723, opened 2019-01-15
Revert "Fix OverflowError in randomPairsMatch()"

fgregg:
Reverts dedupeio/dedupe#722

If `n` is large (which it can easily be), than we will run into a 

```
ValueError: high is out of bounds for uint64 mtrand.pyx:991: ValueError
```

Python ints can be arbitrarily long. 

The previous code may also have been wrong, because when we cast the pure python to a numpy array we may have clipped the values to the max size of the numpy int type. I have to investigate further.

### DIFF
--- a/dedupe/core.py
+++ b/dedupe/core.py
@@ -71,8 +71,8 @@ def randomPairsMatch(n_records_A, n_records_B, sample_size):
     if sample_size >= n:
         random_pairs = numpy.arange(n)
     else:
-        random_pairs = numpy.random.randint(
-            low=1, high=n, size=sample_size, dtype='int64')
+        random_pairs = numpy.array(random.sample(range(n), sample_size),
+                                   dtype=int)
 
     i, j = numpy.unravel_index(random_pairs, (n_records_A, n_records_B))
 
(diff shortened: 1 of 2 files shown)


az0:
You mentioned `n` can cause a ValueError, but is worse with PR722? In other words, was 722 progress?

You mentioned clipping. First, I could switch the numpy `dtype` from `int` to `int64`. Second, I could add a test like this: 

* `n_records_A` is large
* `n_records_B` can be small
* `sample_size` is about 95% of `n_records_A`
* assert that large values of A (i.e., near the value of `n_records_A`) are returned

cmdavis4:
I just ran into the error @az0 originally noted in #722 and which their PR fixed until it was reverted, and I fixed it by manually implementing the changes from their PR. It seems that #780 might be encountering an error similar to the one that was anticipated in the reasoning for reverting this, but their bug report mentions that they also manually implemented the changes from #722 before getting the new error. It seems like working on even moderately sized data can throw the original OverflowError and, as such, that I am probably not the only frustrated user getting that error and then having to comb through bug reports and manually implement a reverted hotfix. I understand wanting a larger refactor, but I would think squashing this simple bug should take precedence.

PR #732, opened 2019-02-10
words -> characters

anjor:
The example actually uses a "first three characters" predicate. This PR fixes the typo in the docs


fgregg:
Thanks!

PR #754, opened 2019-06-20
fix mistake in docs for custom comparator

stevemartingale:
The values returned in the custom comparator example in the docs are the opposite of what they should be.


fgregg:
why do you think that's so?

Issue #751, opened 2019-06-04
Are Mac OS X install notes still relevant?


ghing:
I was trying to understand the notes at https://docs.dedupe.io/en/latest/OSX-Install-Notes.html# and see if they were still relevant in June 2019.

A little Googling indicated that current wheels include openblas (See https://github.com/numpy/numpy/issues/11537) and I confirmed this, I think, on my Mac:

```
pipenv run pip show -f numpy | grep blas
Loading .env environment variables…
numpy/.dylibs/libopenblasp-r0.3.7.dev.dylib
```

Are the installation steps in the documentation still needed?

fgregg:
do you see an warning like these?

https://github.com/dedupeio/dedupe/blob/053d373aaed47201f720c5b6d1a568fc49742cc3/dedupe/backport.py#L19-L26

ghing:
I don't remember seeing those. Would those have been suppressed if I used `csvlink`? At what stage of the process would I have seen them?

ghing:
I looked a bit closer at this module and replicated the checks in it.  Here's what the `config_info` list looks like for my  install which got installed automatically when I installed pandas in my virtualenv.

```
>>> import numpy
>>> from future.utils import viewitems
>>> config_info = str([value for key, value in
                   viewitems(numpy.__config__.__dict__)
                   if key.endswith("_info")]).lower()
>>> config_info
"[{}, {}, {'libraries': ['openblas', 'openblas'], 'library_dirs': ['/usr/local/lib'], 'language': 'c', 'define_macros': [('have_cblas', none)]}, {'libraries': ['openblas', 'openblas'], 'library_dirs': ['/usr/local/lib'], 'language': 'c', 'define_macros': [('have_cblas', none)]}, {}, {'libraries': ['openblas', 'openblas'], 'library_dirs': ['/usr/local/lib'], 'language': 'c', 'define_macros': [('have_cblas', none)]}, {'libraries': ['openblas', 'openblas'], 'library_dirs': ['/usr/local/lib'], 'language': 'c', 'define_macros': [('have_cblas', none)]}, <function get_info at 0x1076289d8>]"
```

fgregg:
Great! Would you have time to make a PR?

ghing:
Working on this now.

PR #763, opened 2019-07-23
Document that special OS X steps are no longer needed

ghing:
Current numpy wheels for Mac OS X use OpenBLAS so the special
installation steps are no longer needed.

Continue to link to the special steps for context, but remove it as a
top-level menu item.

Addresses https://github.com/dedupeio/dedupe/issues/751


docs/index.rst
```diff
@@ -70,8 +69,9 @@ Installation
 Mac OS X Install Notes
```
fgregg: thanks, @ghing. Let's be more aggressive and just completely remove this section from this page.

ghing: Ok. Completely removed all references to special instructions in 5281661.

Issue #776, opened 2019-09-20
clustered_dupes is empty when I send the data with smaller_ids


pmalipeddi:
I am using MongoDB to get clusterids and passing the data "records" in the code to deduper.matchBlocks as below but I am not getting any cluster_ids back as  "clustered_dupes" is always empty, am I passing the data to match blocks in a wrong format? if this is  wrong format for match blocks please let me know the correct format
**Records with smaller_ids**:  
[('ETY1211', {'EtyNbr': 'ETY1211', 'block_id': 2, 'smaller_ids': [2], 'Alpha': 'A', 'City': 'EAST HARTFORD', 'ClusterInd': 'N', 'Cntry': 'US', 'Fname': 'A DALE & JUDY', 'FullAddr': '92 LANGFORD LN ', 'Lname': 'HUSTON', 'State': 'CT', 'Zip': '06118', 'cluster_conf': '-1', 'cluster_id': '-1'}, {2}), 
('ETY1213', {'EtyNbr': 'ETY1213', 'block_id': 2, 'smaller_ids': [2], 'Alpha': 'A', 'City': 'EAST HARTFORD', 'ClusterInd': 'N', 'Cntry': 'US', 'Fname': 'A DALE & JUDY', 'FullAddr': '92 LANGFORD LN ', 'Lname': 'HUSTON', 'State': 'CT', 'Zip': '06118', 'cluster_conf': '-1', 'cluster_id': '-1'}, {2})] 

When I pass the data to match blocks as below where with " set()" is empty then I am getting cluster_ids from clustered_dupes
**Records without smaller_ids**:
[('ETY1217', {'EtyNbr': 'ETY1217', 'block_id': 2, 'smaller_ids': [2], 'Alpha': 'A', 'City': 'EAST HARTFORD', 'ClusterInd': 'N', 'Cntry': 'US', 'Fname': 'A DALE & JUDY', 'FullAddr': '92 LANGFORD LN ', 'Lname': 'HUSTON', 'State': 'CT', 'Zip': '06118', 'cluster_conf': '-1', 'cluster_id': '-1'}, set()), 
[...]

Issue #772, opened 2019-08-20
Multiprocessing in later version not working properly


monicatao:
Hi, I upgraded the library and ran gazette but it's getting much slower than the previous versions.  I found it started to get slower since from version 1.9.7. And the function it's slower is dedupe.core.scoreGazette.

I compared the GitHub from 1.9.6 and 1.9.7 and cannot tell the reason why it's happening. 

After some investigation, I think it might be the multiprocessing in the later version is not working properly. I tried it with both multiprocessing and no multiprocessing and running no multiprocessing is even faster than the current one with multiprocessing.

Could you help investigate this? Thank you!!

fgregg:
can you use profiling (https://docs.python.org/3/library/profile.html) to
provide data on the speed difference and where the difference is

fgregg:
this should be fixed in an upcoming release.

michael-clarkson:
Hi @fgregg is this fixed in the 1.10 release?

Thanks!

fgregg:
no, it's fixed on the https://github.com/dedupeio/dedupe/tree/dedupe_2 branch

michael-clarkson:
Ok thanks, do you have a plan about when you would like to release it?

fgregg:
no

monicatao:
Hi @fgregg is there a reason why it's not fixed in the new release but in a different branch?

fgregg:
please feel free to port the fix.

fgregg:
closed by https://github.com/dedupeio/dedupe/pull/788

Issue #680, opened 2018-07-23
Make dedupe easier to use with pandas


fgregg:
Probably won't do this in the core library, but making dedupe work well with pandas could significantly improve adoption in they python data science community. 

I'm not a pandas person, so I'm not sure what it would mean for dedupe to "work well" with pandas. Some notions.

- let people pass in pandas dataframes directly into the dedupe and linking classes
- output a pandas dataframe
- provide an API more similar scikit-learn (I don't know if this is actually helpful to the pandas community, maybe that's a separate issue)
- alternatively to the scikit-learn interface we could use pandas custom accessor interface.

What are other things that would be good for pandas users?

fgregg:
Looks like @Lyonk71 is working on something interesting here: https://github.com/Lyonk71/pandas-dedupe

Lyonk71:
I think that the benefits of connecting Dedupe with Pandas are mostly things like work flow & intuitive design for tabular data. I work with folks in public health who need to deduplicate & link large files. These people tend to be comfortable with SQL, SAS, and R, but experience with Python is not so common. Since these researchers are already comfortable with tables/dataframes & basic programming techniques, I can get them set up with [Ten Minutes to Pandas](https://pandas.pydata.org/pandas-docs/stable/10min.html),  and they can start using pandas-dedupe successfully in about 20-30 minutes.

Even for people with more experience, being able to use Dedupe directly on a dataframe takes away from some of the daily grind of analysis work.

Below is a list of convenience/ease-of-use features I’ve been planning for pandas-dedupe:

•	Limit required code to an import statement, loading a dataframe, and a single line of pandas-dedupe code.

•	Incorporation of SQLite3 for working with larger datasets. A user could set a parameter like low_memory=’yes’, and have SQLite run in the background without any setup, or DB knowledge.

•	Enable fuzzy merging with SQL-style joins

•	Allow users to specify which fields they would like to keep canonical versions for. Use canonical version if available, use original value if no canonical version is available. Delete canonical fields before returning to user so that they get the same number of output fields as input fields.

•	If the user specifies a Dedupe type (String, Price, LatLong, etc.), automatically convert the field to be the correct Python data type, include parameter to remove any rows that cannot be converted.

•	For instances where Dedupe is called multiple times in the same program, provide a parameter to read/write specific .db, settings, and training JSON files.

Hope this helps! 

[...]

Issue #546, opened 2017-04-20
Revisiting the save index methods


fgregg:
Now, the indices that are used in blocking can be saved and restored. This only really seems useful for gazetteer classes. 

With the file-backed blocking introduced in 1.6.8, this becomes a lot more interesting. I think I might reduce the api for the saving and restoring of blocked records and indices for the gazetteer class and remove it from the other classes.

mehrrsaa:
Hi @fgregg,

Using the most recent version of Dedupe, I am still unable to save result of indexAll using writeSettings when working with gazetteer on large data sets and getting the maximum recursion error. Am I missing something here or is this still the case?

Thank you

matthewdchambers:
Hey @fgregg - experiencing the same here. Any guidance would be sincerely appreciated :)

fgregg:
closed by https://github.com/dedupeio/dedupe/pull/788

PR #788, opened 2020-01-06
Dedupe 2

fgregg:
This version makes extensive changes to both make the library more powerful and, I hope, easier to use. It is not backwards compatible.

The biggest change is clarification and simplification of the three classes.

* Rename `Dedupe.match` -> [`Dedupe.partition`](https://docs.dedupe.io/en/dedupe_2/API-documentation.html#dedupe.Dedupe.partition)
* Renaming `RecordLink.match` to [`RecordLink.join`](https://docs.dedupe.io/en/dedupe_2/API-documentation.html#dedupe.RecordLink.join) and having that join have optional different types of constraints
* Removing `match` from `Gazetteer` and only leaving search. The `match` mode we had before is now an option for `RecordLink.join`

At a lower level, we have moved from a block oriented methods to pair oriented modes

`matchBlocks` has been dropped, and replaced by the [`pairs`](https://docs.dedupe.io/en/dedupe_2/API-documentation.html#dedupe.Dedupe.pairs) and [`score`](https://docs.dedupe.io/en/dedupe_2/API-documentation.html#dedupe.Dedupe.score) methods.

This orientation makes for a simper API, in my opinion, but also supports for significant performance improvements.

Using sqlite, we can generate all unique pairs of records to compare in a memory efficient way. This allows us to avoid using the [Kolb Redundant Free Comparison scheme](https://github.com/dedupeio/dedupe/blob/76761d7214df78cdc318621cfb2fba967baacd8b/dedupe/core.py#L118-L125), (which still required a comparison of redundant pairs of records, just cheaply). 

Using sqlite in the library also allows users to deal with record sets in the low millions without 
needing to go to a database.  

Full documentation of the new API here: https://docs.dedupe.io/en/dedupe_2/API-documentation.html#dedupe.Dedupe.score

### Other changes

- Python 3 only
- Static typing and type Hints
- Multiprocessing enabled for Windows
- Multiprocessing mode changed to spawn for Mac OS X
- Moved from CamelCase to lowercase_with_underscore for method names.
- Dropped ability to save indices in save settings. This often not wor, unpredictably
- Renamed Matching.blocker -> Matching.fingerprinter, since that's really what it is doing
- Moved to autodoc for documentation
- Dropped threshold methods. These were never really defensible.

### Related issues

- #798, #789, #776, #772, #546, #483, #395, #680 

## TODO
- [x] update dedupe-examples
- [x] update csvdedupe, (specify max version, for now)
- [x] add docs for console_label
- [x] adjust docs for fingerprinting vs blocking
- [x] document threshold methods of many_to_n


fgregg:
dedupe-examples have been updated. 

PR here: https://github.com/dedupeio/dedupe-examples/pull/105
Branch here: https://github.com/dedupeio/dedupe-examples/tree/dedupe_2

fgregg:
attn @Lyonk71, @maxharlow, @fjsj, 

I'm about a week away from bringing in these changes. If wanted you to be aware so you can adjust downstream and give any feedback you might like.

PR #809, opened 2020-04-18
Closes #808: fix scoring compatibility with classifiers that return 0 scores

### LINKED ISSUE #808: ValueError "could not broadcast input array" during scoring, new code has wrong behavior
Hi, I'm getting the following exception on Dedupe 2.0.0:

```
.../venv/lib/python3.7/site-packages/dedupe/core.py in fieldDistance(self, record_pairs)
    158 
--> 159                 scored_pairs['pairs'] = ids
    160                 scored_pairs['score'] = scores

ValueError: could not broadcast input array from shape (249,2) into shape (228,2)
```

Looks like the new code is wrong, the old code used a mask:
https://github.com/dedupeio/dedupe/blob/96f9b895948f856637591c4b3dc9807eadf0d626/dedupe/core.py#L143-L148

The new code doesn't:
https://github.com/dedupeio/dedupe/blob/7df2ab5ef268080a60e1f3f31248de985d93c598/dedupe/core.py#L155-L157

That's problematic because the new code is still using `numpy.count_nonzero`. So the `scored_pairs` array may get a smaller shape than `ids`, causing the broadcast error. That happen when `scores` contain zeros.

Please let me know if a full reproduction is useful or if this is already clear enough.

fjsj:
Closes #808

### DIFF
--- a/dedupe/core.py
+++ b/dedupe/core.py
@@ -142,7 +142,8 @@ class ScoreDupes(object):
             distances = self.data_model.distances(records)
             scores = self.classifier.predict_proba(distances)[:, -1]
 
-            if scores.any():
+            mask = scores > 0
+            if mask.any():
                 id_type = sniff_id_type(record_ids)
                 ids = numpy.array(record_ids, dtype=id_type)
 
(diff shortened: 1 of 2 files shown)


dedupe/core.py
```diff
@@ -304,7 +305,7 @@ def __call__(self, block: RecordPairs) -> numpy.ndarray:
         dtype = numpy.dtype([('pairs', id_type, 2),
                              ('score', 'f4')])
 
-        scored_pairs = numpy.empty(shape=numpy.count_nonzero(scores),
+        scored_pairs = numpy.empty(shape=scores.shape[0],
```
fjsj, line 308: Please test this manually, I think this code is untested on pytest, right?

tests/test_core.py
```diff
@@ -96,6 +96,30 @@ def test_score_duplicates(self):
         numpy.testing.assert_allclose(scores['score'],
                                       self.desired_scored_pairs['score'], 2)
 
+    def test_score_duplicates_with_zeros(self):
+        self.classifier.weights = [-1000]
+        self.classifier.bias = 1000
```
fjsj, line 101: Forces classifier to return 0 score.

Issue #819, opened 2020-05-20
disk based connected components


fgregg:
The current scaling bottleneck is usually the connected components step in clustering. The current algorithm loads the entire edgelist into memory. 

We need to move this to disk. 

Things worth exploring

1. connected components in sqlite https://stackoverflow.com/questions/33465859/a-number-of-connected-components-of-a-graph-in-sql

2. look for or implement a specialized, lightweight graph store like https://link.springer.com/chapter/10.1007/978-3-319-01863-8_20

fgregg:
cc @fjsj, perhaps relevant to your interests.

fgregg:
really interesting idea:

https://twitter.com/mccastelle/status/1263272272787320840

fgregg:
https://arxiv.org/abs/1802.09478

fjsj:
Thanks @fgregg, I'll investigate.

With current implementation, I had memory errors at this line, when numpy needs to materialize part of the mmap-ed `scored_pairs/edgelist` array into main memory because of the fancy indexing:

https://github.com/dedupeio/dedupe/blob/20736bb693aeffa0ac6fb1990485fb39f60c0d7b/dedupe/clustering.py#L36

I'm not 100% sure, but I think numpy is smart enough to not load the whole `scored_pairs` into main memory up to that line. That line is where things break for me. Therefore, do any of [those papers](http://poloclub.gatech.edu/mmap/) describe a way of running connected components without materializing a large component in main memory?

Also, worth noting most of my problems were solved when I stopped using non-integer IDs. This dramatically reduced the size of `scored_pairs`. Maybe it's worth adding a warning when `sniff_id_type` detects a non-integer ID type. Please LMK if this is something you agree with so I can make a small PR.

fgregg:
thanks that's very helpful.

fgregg:
@fjsj i think i have some ideas of what would help here. would it possible for you to send me an edgelist for job that ran into memory problems (`scored_pairs['pairs']`)

fjsj:
@fgregg Good to know! I can't share it because it's private data, but I'll try to generate one.

fgregg:
don't need the underlying data, just the id pairs.

fgregg:
I've addressed the memmap issue, but much of union-find is still in-memory. Le's see how much headroom the memmap fix buys us.

fgregg:
superseded by #826

fgregg:
[...]

PR #822, opened 2020-05-24
keep in memmap

fgregg:
This PR adjusts connected components so as to use get the components of by means of slices of the the memmaped array. This keeps the array mainly on disk instead of in memory.

There is still an fundamental memory limit in the approach we are taking (parts of the union_find algorithm are still in-memory), but this should buy us some room.

Figure_1
Figure_2
Figure_3


fgregg:
related to #819

PR #834, opened 2020-06-25
filtered out pairs with zero scores in partition

### LINKED ISSUE #833: infinite loop in clustering.py _connected_components due to zero pair scores
Hi, 

Encountered an issue where the _conencted_components goes into an infinite loop (see screenshot below). I think this occurs when there are pairs that have zero scores. Since numpy treats log of 0 as -Inf, the calculated threshold becomes 0 and the cycle repeats forever. We can remedy this by filtering pair_scores to have scores only above 0. What do you think?

image

Edit: it's not actually infinite looping, but there is unnecessary looping for each additional zero score in the component. Also, the zero scores don't contribute anything to the clustering step, so it's just extra wasted memory space.

joshuacortez:
resolves #833

### DIFF
--- a/dedupe/clustering.py
+++ b/dedupe/clustering.py
@@ -74,6 +74,9 @@ def _connected_components(edgelist: numpy.ndarray,
 
         if n_components > max_components:
             min_score = numpy.min(sub_graph['score'])
+            if min_score == 0:
+                sub_graph = sub_graph[sub_graph['score'] > 0]
+                min_score = numpy.min(sub_graph['score'])
             min_score_logit = numpy.log(min_score) - numpy.log(1 - min_score)
             threshold = 1 / (1 + numpy.exp(-min_score_logit - 1))
             logger.warning('A component contained %s elements. '
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -168,7 +168,6 @@ class DedupeMatching(IntegralMatching):
         """
         pairs = self.pairs(data)
         pair_scores = self.score(pairs)
-        pair_scores = pair_scores[pair_scores["score"] > 0]
         clusters = self.cluster(pair_scores, threshold)
 
         clusters = self._add_singletons(data, clusters)
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -168,6 +168,7 @@ class DedupeMatching(IntegralMatching):
         """
         pairs = self.pairs(data)
         pair_scores = self.score(pairs)
+        pair_scores = pair_scores[pair_scores["score"] > 0]
         clusters = self.cluster(pair_scores, threshold)
 
         clusters = self._add_singletons(data, clusters)


fgregg:
hi @joshuacortez. thanks for your PR. Could you explain the benefits of this change.

joshuacortez:
hi @fgregg. It solves the infinite loop issue during the clustering step in _connected_components function in clustering.py. 

Relevant part of the source code starting here:
https://github.com/dedupeio/dedupe/blob/8e07ece14f4187de25427f05fa775c2be0528f62/dedupe/clustering.py#L75-L96

If there is a sufficiently large component with a zero score, the min_score = 0.
Then, min_score_logit = numpy.log(0) - numpy.log(1 - 0) = -inf
Then, threshold = 1 / (1 + numpy.exp(inf- 1)) = 0

The threshold becomes zero again, resulting in an infinite loop. 
Have encountered this issue myself with these warnings


Apart from this infinite loop issue, I also believe the pair scores with zeros won't contribute to the clustering step. It's just extra space. We can save on storage/memory by discarding them.

joshuacortez:
Got it, pushed a new commit that filters out the scores in _connected_components directly.

I just have a related question to the zero looping issue and I would like to hear your thoughts on this. I, with a coworker, tried deduping a 13 million record dataset on a 64GB RAM machine. We were successful until the clustering step. We initially had the zero looping problem, but even after addressing that, we were still running into out of memory errors.
image

We plotted the histogram of probability scores of record pairs (these are the pairs after blocking) to investigate. 
image

I saw that many probability scores were mostly near zero anyway, and filtering them can solve the out of memory error without affecting the deduping results. So, I set a `hard_threshold` of 0.1. I filtered the `pair_scores` before it gets passed in the `cluster` method. (This is the relevant bit if in the partition function)
https://github.com/dedupeio/dedupe/blob/5092f2598f1f2bcb4f3b1f0ec85f88b95ce7454b/dedupe/api.py#L169-L171

It could be something like:
```python
pairs = self.pairs(data)
pair_scores = self.score(pairs)
pair_scores = pair_scores[pair_scores['score'] > hard_threshold]
clusters = self.cluster(pair_scores, threshold)
```

After setting the `hard_threshold`, everything worked smoothly, with no out of memory error.

My question is, do you think it is worth filtering the `pair_scores` via a `hard_threshold` before passing it in the `cluster` method? Are the near-zero scores still worth using for the clustering step? I'm not sure if the histogram of probability scores looks different in other use cases of the Dedupe library. But in my experience, in running it for other datasets, there's typically a large proportion of near-zero scores.

fgregg:
this is a very keen observation. 

i agree that it is typical for most of the scored comparisons to be close to zero **and** that for many values of the clustering threshold, we could drop many of these near-zeros without impacting the final clusters

what's not clear to me is how to use that insight.

I'd rather not add a `hard_threshold` arg to the API, because i think it would be hard to explain and right now we can't really provide any guidance on how to use it.

the ideal would be to have a `pair_threshold` be a function of the clustering threshold. the magic function would tell us that, given a clustering threshold, what is a pair score threshold such that we can ignore any pairs that have score less than threshold without changing the final clustering.

i'm not sure if such a function exists, but studying the centroid clustering criterion we use. https://web.stanford.edu/class/cs345a/slides/12-clustering.pdf

dedupe/clustering.py
```diff
@@ -74,6 +74,9 @@ def _connected_components(edgelist: numpy.ndarray,
 
         if n_components > max_components:
             min_score = numpy.min(sub_graph['score'])
+            if min_score == 0:
```
fjsj, line 77: After https://github.com/dedupeio/dedupe/pull/809 was merged, I think we shouldn't see `0` scores at clustering stage. Does your code have those changes?

fgregg:
What we would like to do is find a value such that we can zero out scores below that value without significant changes in the final clustering.

Assume we are using the [average method of clustering linkage](https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html).

Assume we have an existing cluster of records and record outside the clsuter. Every record in this cluster has a record pair score to the outside record. Let one of these record pairs have value of `y` and all other pairs have value `z`.

The cophenetic distance of adding this additional record to the cluster is

`(y + k*z)/(k + 1)` where `k + 1` is the number of records already in the cluster.

Let `y=0` then the size of the cluster `k` that we need for the distance to fall within the clustering threshold `t` is 

`k =  -t/(t - z)`

let y be some value larger than 0, then the number of records `l` that we need for the cluster to fall within the clustering threshold is 

`l = (y - t)/(t - z)`

what is the value of `y` such that replacing that value with 0 will very small effect on the clustering results.

One way to operationalize "very small effect" is to limit how much larger `k` is than `l`. 

As a first pass, we could say that 

`k - l < 1`

but this doesn't really capture scale.

Instead let's say that

`(k - l)/k < 1/100`

in order achieve that inequality, we can simply have `y = t/100`.

We can zero out values (i.e. filter them out record pairs) below this threshold and it should have a minimal effect.

@joshuacortez, @fjsj does this make as a criterion.

(Note that we actually use the centroid method for record linkage, but it should behave pretty similarly)

fjsj:
@fgregg This sounds nice. But how to decide to use `/100` or a larger denominator? Maybe the best denominator depends on the average/median cluster size?

Also, your proposal seems similar to "global threshold sparsification". There are other sparsification methods that are local. See: https://www.researchgate.net/publication/225919431_Unsupervised_Sparsification_of_Similarity_Graphs and https://www.researchgate.net/publication/328034827_Improving_Short_Text_Clustering_by_Similarity_Matrix_Sparsification

fgregg:
i'm going to close this for now. 

@fjsj, if you wanted to propose a sparsification method based on your research, i would be obliged.

PR #837, opened 2020-07-13
Closes #836: escape colon at CompoundPredicate

### LINKED ISSUE #836: CompoundPredicate should escape `:` to avoid wrong block_keys
I've found an issue on `CompoundPredicate` caused by the lack of escaping `:` chars inside fields.

This causes two problems. First, it can generate duplicate `block_key`s for the same `id`. Here's the example:

```
(SimplePredicate: (commonSetElementPredicate, name), SimplePredicate: (commonSetElementPredicate, address))
```

```
id  | name            | address    
---------------------------------------
1   | {'foo:', 'foo'} | {':bar', 'bar'}
```

```
id | block_key
---------------
1  | foo::bar:0
1  | foo::bar:0
```

That's probably not problematic for Dedupe 2, but that breaks [this line](https://github.com/dedupeio/dedupe-examples/blob/83dbf872674a5c3f6a209c367f46c7e6ef78e9a3/pgsql_big_dedupe_example/pgsql_big_dedupe_example.py#L246-L247) of the old `smaller_id`-based big Dedupe I still use.

Second, it can group together under the same `block_key` records that shouldn't be blocked together. See:

```
(SimplePredicate: (wholeFieldPredicate, name), SimplePredicate: (wholeFieldPredicate, address))
```

```
id  | name    | address    
---------------------------
1   | flavio: | one street
2   | flavio  | :one street
```

```
id | blocking_key
-------------------------
1  | flavio::one street:0
2  | flavio::one street:0
```

The solution is to escape `:` right before performing the `join` at this line:

https://github.com/dedupeio/dedupe/blob/5092f2598f1f2bcb4f3b1f0ec85f88b95ce7454b/dedupe/predicates.py#L332

Something like `u':'.join(block_key.replace(u':', u'\\:'))` would work, because the examples above would become this:

```
id | block_key
---------------
1  | foo\::bar:0
1  | foo:\:bar:0
```

```
id | blocking_key
-------------------------
1  | flavio\::one street:0
2  | flavio:\:one street:0
```

I'll open a PR for this.

fjsj:
Closes #836

### DIFF
--- a/dedupe/predicates.py
+++ b/dedupe/predicates.py
@@ -329,9 +329,14 @@ class CompoundPredicate(tuple):
     def __call__(self, record, **kwargs):
         predicate_keys = [predicate(record, **kwargs)
                           for predicate in self]
-        return [u':'.join(block_key)
-                for block_key
-                in itertools.product(*predicate_keys)]
+        return [
+            u':'.join(
+                # must escape : to avoid confusion with : join separator
+                b.replace(u':', u'\\:') for b in block_key
+            )
+            for block_key
+            in itertools.product(*predicate_keys)
+        ]
 
 
 def wholeFieldPredicate(field: Any) -> Tuple[str]:
--- a/tests/test_predicates.py
+++ b/tests/test_predicates.py
@@ -80,5 +80,26 @@ class TestNumericPredicates(unittest.TestCase):
         assert predicates.roundTo1(-22315) == (u'-20000',)
 
 
+class TestCompoundPredicate(unittest.TestCase):
+    def test_escapes_colon(self):
+        '''
+        Regression test for issue #836
+        '''
+        predicate_1 = predicates.SimplePredicate(
+            predicates.commonSetElementPredicate, 'col_1')
+        predicate_2 = predicates.SimplePredicate(
+            predicates.commonSetElementPredicate, 'col_2')
+        record = {
+            'col_1': ['foo:', 'foo'],
+            'col_2': [':bar', 'bar']
+        }
+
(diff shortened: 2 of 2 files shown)


fgregg:
Doesn't this just push the problem to even more esoteric blocks that have the `\:` string w/i them? Perhaps we want to use something like an ASCII unit separator? https://en.wikipedia.org/wiki/C0_and_C1_control_codes#Field_separators

fjsj:
@fgregg I don't think it pushes the problem, because `\:` will also be escaped. Added a test to demonstrate that.
AFAIK, for proper escaping, you must replace with two characters with one of them being the character you want to escape. So I don't think the solution is esoteric. It's how Python does escaping on CSV files for example:

```python
In [1]: import csv, io 
   ...: f = io.StringIO() 
   ...: w = csv.writer(f, escapechar='\\', delimiter=':', quoting=csv.QUOTE_NONE) 
   ...: w.writerow(['foo:', ':bar']) 
   ...: f.getvalue()                                                                                                         
Out[1]: 'foo\\::\\:bar\r\n'
```

> Perhaps we want to use something like an ASCII unit separator?

Yes, that's another option. I opted to not use it to facilitate block keys debugging, especially on DB queries. But I can change the PR to use that if you prefer.

fgregg:
sorry, didn't meant to imply that your solution was esoteric, just that it would be pretty rare to have strings with "\:" in them.

Thanks for the test! looks good, and agree it's a better solution than a hidden character.

fjsj:
No problem, I misunderstood. Thanks!

PR #854, opened 2020-09-04
Random Forest Inspired Predicate Candidates

fgregg:
Our current block learner proceeds as follows

1. Calculate the coverage and comparison cost for all compound predicates of length 1 and 2
2. For this set of predicates, use Branch and Bound to find the subset that covers all the co-referent pairs and has the lowest total comparisons

It would be good to consider compound predicates of length 3 and higher, but this is not feasible to use our current procedure because of the combinatorial explosion.

This PR takes an approach inspired by random forests.

1. take a subsample of co-referent records and subsample of simple predicates, greedily find the best predicate of length k for these values
2. repeat sampling to produce candidate set
3. use Branch and Bound with that candidate set to find good subset.

many thanks to @tarakc02 for thinking through this with me.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -4,7 +4,7 @@ on: [push]
 
 jobs:
   test:
-
+    timeout-minutes: 5
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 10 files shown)


fgregg:
@fjsj, this is going very nicely. it's having 3+ conjunctions is likely to substantially increase the amount of time we spend blocking, so this work might make your work on parallelizing more urgent.

fjsj:
@fgregg sounds good, I'll prioritize the parallel blocking work.

fgregg:
don't have "all predicates" be the starting point for matching

PR #864, opened 2020-11-07
Don't raise AssertionError when out of candidates

alextsits:
Fix a bug where AssertionError was raised indicating that an instance
should be initialized with `sample()`, even though the instance is
already initialized. The problem is that some learners (like the
`DisagreementLearner`) define the `__len__()` method, returning the
number of candidates. However this means that the assertion could fail
if the learner ran out of candidates. Fix that by asserting explicitly
that the variable is not None.

Also, since `uncertain_pairs()` can raise IndexError when no candidates
exist, handle this case in `console_label()` too, just like we already
do when we `pop()` from `uncertain_pairs`.

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -1037,7 +1037,8 @@ class ActiveMatching(Matching):
                               and take substantial memory.
 
         """
-        assert self.active_learner, "Please initialize with the sample method"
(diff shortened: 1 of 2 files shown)


fgregg:
thanks! great catches.

PR #880, opened 2021-03-08
implements `return_confidence_scores` as argument to clustering

fritshermans:
We found that the confidence score calculation was very time consuming when using dedupe with large data. This pull request implements a `return_confidence_scores` argument in the cluster function that defaults to True. It allows the user to set it to False to gain significant speed improvements when confidence scores are not required.

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -256,15 +256,16 @@ class DedupeMatching(IntegralMatching):
             pairs.close()
             con.close()
 
-    def cluster(self,
-                scores: numpy.ndarray,
-                threshold: float = 0.5) -> Clusters:
+    def cluster(self, scores: numpy.ndarray,
+                threshold: float = 0.5,
+                return_confidence_scores: bool = True) -> Clusters:
         r"""From the similarity scores of pairs of records, decide which groups
         of records are all referring to the same entity.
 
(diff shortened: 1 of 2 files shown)


fgregg:
that's interesting! i like the idea of calculating score being optional. i am not sure about the API yet.

fritshermans:
I tested it and it works well. In one case it decreased computation time from 9 hours to 14 minutes :-). Could you let me know what you mean with you comment about the API? Let me know if you want me to test or add anything.

fgregg:
that's a remarkable difference. could you possibly provide an edgelist where the caclulating the scores makes such a noticeable difference?

fritshermans:
We work with very large datasets containing company names and addresses. I would not be allowed to share this data. We found that the connected components algorithm sometimes returns a very large component that causes these extreme calculation times. However, by disabling the confidence score calculation the clustering time is significantly reduced.

fgregg:
sure, i was hoping that the edgelist that's input into the clustering function would be sufficiently anonymized that it could be shared.

i think my first preference would be to see if we could speed up the implementation of calculating the confidence scores, and i'll need some data to work with for that.

fgregg:
#912 complete addresses the speed problems here, so I will close this.

PR #886, opened 2021-03-18
implements firstTwoTokensPredicate

fritshermans:
I found that often the two words of a company or person name are equal and therefore this appeared to be a good blocking rule. Since it is more restrictive than the firstTokenPredicate, the number of pairwise comparisons that needs to be made are reduced drastically when this predicate is selected.

### DIFF
--- a/dedupe/predicates.py
+++ b/dedupe/predicates.py
@@ -18,7 +18,7 @@ from dedupe._typing import RecordDict
 words = re.compile(r"[\w']+").findall
 integers = re.compile(r"\d+").findall
 start_word = re.compile(r"^([\w']+)").match
-two_start_words = re.compile(r"^[\w']+(\s+[\w']+)?").match
+two_start_words = re.compile(r"^([\w']+\s+[\w']+)").match
 start_integer = re.compile(r"^(\d+)").match
 alpha_numeric = re.compile(r"(?=\w*\d)[a-zA-Z\d]+").findall
 
@@ -358,7 +358,7 @@ def firstTokenPredicate(field: str) -> Sequence[str]:
 def firstTwoTokensPredicate(field: str) -> Sequence[str]:
     first_two_tokens = two_start_words(field)
     if first_two_tokens:
-        return (first_two_tokens.group(),)
+        return first_two_tokens.groups()
     else:
         return ()
 
--- a/dedupe/predicates.py
+++ b/dedupe/predicates.py
@@ -18,6 +18,7 @@ from dedupe._typing import RecordDict
 words = re.compile(r"[\w']+").findall
 integers = re.compile(r"\d+").findall
 start_word = re.compile(r"^([\w']+)").match
+two_start_words = re.compile(r"^[\w']+(\s+[\w']+)?").match
 start_integer = re.compile(r"^(\d+)").match
 alpha_numeric = re.compile(r"(?=\w*\d)[a-zA-Z\d]+").findall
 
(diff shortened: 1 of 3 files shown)


dedupe/predicates.py
```diff
@@ -354,6 +355,14 @@ def firstTokenPredicate(field: str) -> Sequence[str]:
         return ()
 
 
+def firstTwoTokensPredicate(field: str) -> Sequence[str]:
+    first_two_tokens = two_start_words(field)
+    if first_two_tokens:
+        return (first_two_tokens.group(),)
```
fgregg, line 361: can this just be?

```
if first_tokens:
    return first_two_tokens.groups()
```

fgregg (changes requested):

fritshermans:
Unfortunately that's not possible as there needs to be a capturing group in the regex for the second word (with thee '?'  quantifier).

dedupe/predicates.py
fritshermans: Unfortunately that's not possible as there needs to be a capturing group in the regex for the second word (with thee '?' quantifier).

tests/test_dedupe.py
```diff
@@ -236,6 +236,9 @@ def test_predicates_correctness(self):
         assert dedupe.predicates.firstTokenPredicate(field) == ('123',)
         assert dedupe.predicates.firstTokenPredicate('') == ()
         assert dedupe.predicates.firstTokenPredicate('123/') == ('123',)
+        assert dedupe.predicates.firstTwoTokensPredicate(field) == ('123 16th',)
+        assert dedupe.predicates.firstTwoTokensPredicate('') == ()
+        assert dedupe.predicates.firstTwoTokensPredicate('123/') == ('123',)
```
fgregg, line 241: why should this not return an empty tuple?

fgregg: to me, a predicate called first two tokens should return the first two tokens, and if there are not two tokens then it should return nothing

fritshermans: Hi Forest, I thought about this and my reasoning was as follows. Say we have the following list of company names:

- ING Bank
- ING Bank NV
- ING Bank Group
- ING
- Ing Janssen
- Ing Smith

I apologize for the silly example, however it is a good illustration of what we see in real company name data.

In this list, the first four should be deduplicated as one real world entity and the last two should not be deduplicated but remain as two singletons. The predicate I propose would result in 6 pairs: (ING Bank, ING Bank NV), ..., (ING Bank Group, ING). If the predicate would not return the word in case of a single word string, the `firstTokenPredicate` would be needed resulting in 15 pairs for this example. 

That's why I propose to make the new predicate return a single word in case of a one word string. We could rename the predicate obviously to make more clear what it does. 

Does this make sense? I'm curious on your thoughts as you obviously thought about these things
much longer.

fgregg: i understand your point. I'd like to explore if your pattern makes more sense, but right now this is not how equivalent existing predicates work. 

if you are interested in pursuing this pattern, then i would advise

1. in this PR, change the behavior such that if there is only one token, then the predicate returns an empty tuple
2. in a second PR, change the behavior for all the predicates of this pattern (sameThreeCharStart, sameFiveCharStart, sameSevenCharStart, commonTwoElement, commonThreeElement, commonTwoTokens, commonThreeTokens, commonFourGram, commonSixGram)

fritshermans: Sure, i'll do that.

fgregg:
thanks!

PR #909, opened 2021-07-21
Remove recursion in setup.py

Scriptkiddi:
dedupe-variable-datetime is listed as a requirement in setup.py in dedupe, but dedupe is listed in the dedupe-variable-datetime setup.py as well.
This results in a cyclic dependency. As I have found no code usages of dedupe-variable-datetime in dedupe, I would suggest removing it from the setup.py.


fgregg:
it’s a namespace package that makes a date time variable available to the user

Issue #911, opened 2021-08-20
Squared Distances Recalculated Unnecessarily 


jhendr:
In [clustering.confidences](https://github.com/dedupeio/dedupe/blob/master/dedupe/clustering.py#L250) squared_distances is computed each time by squaring condensed_distances. However, confidences is called in a [loop](https://github.com/dedupeio/dedupe/blob/master/dedupe/clustering.py#L250) in which condensed_distances is unchanging. In the case where condensed_distances is large and there are a large number of small clusters (meaning many loop iterations) this can result in repeating that computation many times. In my use case squaring condensed_distances takes around 500ms, but is repeated around 50k times wasting hours of time.

This may be some or all of the problem described in https://github.com/dedupeio/dedupe/pull/880.

Will post a PR shortly...

fgregg:
closed by #912

PR #912, opened 2021-08-20
pre-compute squared_distances

jhendr:
Addresses https://github.com/dedupeio/dedupe/issues/911 and possibly related to https://github.com/dedupeio/dedupe/pull/880.

Pre-computing squared distances avoids it being recomputed in a loop each time `confidences` is called. This saves a lot of time when the loop has a large number of iterations and condensed distances is large.

### DIFF
--- a/dedupe/clustering.py
+++ b/dedupe/clustering.py
@@ -212,6 +212,7 @@ def cluster(dupes: numpy.ndarray,
         if len(sub_graph) > 1:
 
             i_to_id, condensed_distances, N = condensedDistance(sub_graph)
+            squared_distances = condensed_distances**2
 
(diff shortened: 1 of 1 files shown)


dedupe/clustering.py
```diff
@@ -228,7 +229,7 @@ def cluster(dupes: numpy.ndarray,
```
fgregg, line 229: great observation @jhendr, let's move the created of squared_distances down here, closer to where use it.

fgregg (changes requested):

fgregg:
you may need to merge from master to fix the mypy complaints

dedupe/clustering.py
jhendr: 👍  Done

PR #913, opened 2021-08-20
add in-memory option

jhendr:
Using sqlite for join operations is good when memory is an issue, but has a performance hit when it is not. Depending on the input data, parameterization of the model, and machine running the code it may be preferable to run in memory or use sqlite to push things to disk. I'm working with a large dataset, but on a machine with plenty of RAM.

Conveniently sqlite3 gives the option to run in memory simply by changing the connection string to `:memory:`. See [here](https://docs.python.org/3/library/sqlite3.html). In my particular case changing the connection from a temporary file to working in memory reduces compute time of partition from 91 minutes to 24 minutes.

I propose adding an `in_memory` option that defaults to False (maintaining existing behavior) but gives the option to run things in memory. Thanks to the sqlite3 option, it comes with very little additional complexity.

If you are interested in this change maybe it makes sense to add test cases, but I'm not sure where they belong exactly. I've tried running tests/canonical_matching.py and  tests/canonical.py with in_memory hard coded to True and False to verify that much works.

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -61,13 +61,17 @@ class Matching(object):
     Base Class for Record Matching Classes
     """
 
-    def __init__(self, num_cores: Optional[int], **kwargs) -> None:
+    def __init__(self, 
+                 num_cores: Optional[int],
+                 in_memory: bool = False,
+                 **kwargs) -> None:
 
         if num_cores is None:
             self.num_cores = multiprocessing.cpu_count()
         else:
             self.num_cores = num_cores
 
+        self.in_memory = in_memory
         self._fingerprinter: Optional[blocking.Fingerprinter] = None
         self.data_model: datamodel.DataModel
         self.classifier: Classifier
@@ -128,8 +132,7 @@ class DedupeMatching(IntegralMatching):
 
     def partition(self,
                   data: Data,
-                  threshold: float = 0.5,
-                  in_memory: bool = False) -> Clusters:  # pragma: no cover
+                  threshold: float = 0.5) -> Clusters:  # pragma: no cover
         """
         Identifies records that all refer to the same entity, returns
         tuples containing a sequence of record ids and corresponding
@@ -157,9 +160,6 @@ class DedupeMatching(IntegralMatching):
 
                        Lowering the number will increase recall,
                        raising it will increase precision
-                    
-            in_memory: Boolean that if True will compute pairs using
-                       sqlite in RAM rather than writing to disk.
 
         .. code:: python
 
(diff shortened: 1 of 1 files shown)


dedupe/api.py
```diff
@@ -227,7 +233,10 @@ def pairs(self, data):
         # Blocking and pair generation are typically the first memory
         # bottlenecks, so we'll use sqlite3 to avoid doing them in memory
         with tempfile.TemporaryDirectory() as temp_dir:
-            con = sqlite3.connect(temp_dir + '/blocks.db')
+            if in_memory:
+                con = sqlite3.connect(':memory:')
```
jhendr, line 237: It doesn't seem like the worst thing, but it is a bit awkward that the temp_dir still gets created in this case where it is unused. It looks like there are some nice patterns for avoiding this, but some of them depend on python version >= 3.3 or 3.7. I'm not sure what dedupe is trying to support currently. See some of the answers on this stackoverflow: https://stackoverflow.com/questions/27803059/conditional-with-statement-in-python.

May not be worth worrying about the temp_dir at all.

fgregg:
i'm a bit surprised that you had a dataset that was large enough that the job ran for 90 minutes but where the blocking map wasn't enormous. Or maybe it was very big, but you have lots of memory on your machine?

In general, I'm open to this option. I think it should be a config option when initializing the Dedupe, RecordLink, or Gazetteer class, not an argument to the various methods.

fgregg:
you may need to merge from master to fix the mypy complaints

jhendr:
Made changes that move the in_memory option to a class attribute determined on initialization. I have it sitting in the parent `Matching` class similar to num_cores. Let me know what you think.

I am usually running on a machine with very large RAM for my primary use case. That said, I'm profiling on my laptop and not seeing all that much memory usage. Like you mention, I've been a little surprised by the combination of runtime, size of input data, memory usage, etc. There are some things about the performance characteristics of dedupe I'm having a hard time nailing down, but I'm doing some digging to see what I can find, hence this and my other PR.

fgregg:
can you add documentation for these arguments here https://github.com/dedupeio/dedupe/blob/master/docs/API-documentation.rst

jhendr:
> can you add documentation for these arguments here https://github.com/dedupeio/dedupe/blob/master/docs/API-documentation.rst

Looks like that rst file is mostly using sphinx to pick things up automatically from docstrings. Would you rather I add directly to the rst or just put it in the relevant docstring? in the `Matching` class?

fgregg:
i think if you update the docstring in the ActiveMatching and StaticMatching classes, it will do the right thing, but i don't completely recall.

jhendr:
image

Looks like you are correct. That seems to work. Since the documentation lives in ActiveMatching and StaticMatching, I added `in_memory` as an explicit kwarg in those classes rather than just having it pass implicitly through **kwargs.

fgregg (approved):

fgregg:
very nice, thanks a lot

PR #938, opened 2022-01-24
Make use of "Index" and "Canopy" in docs consistent

NickCrews:
I don't really understand the difference between Index
and Canopy blocks (Index blocks are a type of blocks,
and canopy clustering is an algorithm to implement them?),
but I was getting a bit confused from reading the docs as to
the difference between them, and how they relate to the
`index_predicates` argument to the `train()` meethods.

So, this is my best guess as to how to make their use in
the docs more consistent. I would be happy to improve
this if you point me in the right direction of what to read
up on.


NickCrews:
Replaced by #939

PR #939, opened 2022-01-25
Docs improvements

NickCrews:
Various small improvements to the docs that I think I would have appreciated when I first started using this library.

Replacement for #938


fgregg:
nice improvements. much appreciated.

PR #944, opened 2022-01-26
Add Docs build to CI

NickCrews:
There was a bug in that last PR that was merged, such that a reference didn't work. This could have gotten caught in CI, so this adds building the docs to the CI run.

This also fixes up some warnings and errors that previously existed but were slipping through. You can see the logs of the failed docs builds by looking at the Github Action runs for each of the commits in  https://github.com/NickCrews/dedupe/commits/docs-CI

IDK exactly how the docs are built in prod (I assume its readthedocs?) but you could also solve this problem using the readthedocs GitHub integration (https://docs.readthedocs.io/en/stable/integrations.html), so that RTD actually does the build and you get a more "true" preview of the final product.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -W
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = _build
--- a/docs/API-documentation.rst
+++ b/docs/API-documentation.rst
@@ -136,6 +136,7 @@ Dedupe and StaticDedupe
 .. currentmodule:: dedupe
 
 .. class:: Dedupe
+   :noindex:
 
     .. attribute:: fingerprinter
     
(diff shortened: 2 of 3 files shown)


docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -W
```
NickCrews, line 5: This flag makes warnings be treated as errors. Should have put this info int he commit msg

fgregg:
thanks! I turned on the rtd integration

fgregg:
@NickCrews could you pull out the GH changes, because it looks like there are also good doc changes in this PR?

NickCrews:
Sounds good with the RTD integration, that's a better solution. But you still probably want to
- pull in the second two commits that actually fix the docs
- Still turn on the -W flag so that the RTD build fails on warnings

NickCrews:
> @NickCrews could you pull out the GH Action changes, because it looks like there are also good doc changes in this PR?

You want me to keep in adding the -W flag? Without that, the RTD build wouldn't have caught the Fingerprinter.index_fields bug

fgregg:
yeah, let's keep the flag!

fgregg:
thanks so much, @NickCrews

Issue #948, opened 2022-02-01
Add "How-to guide" style docs


NickCrews:
Following the model of the "[Four kinds of documentation](https://documentation.divio.com/)", the docs are currently great in many regards
- Great API docs ("Reference")
- Great reference for variable definitions ("Reference")
- Fantastic "How it works" for background knowledge ("Explanation")
- Good examples ("Walkthrough")

But, I think the docs are missing the 4th type, the "How-to guide". An example of a good version of this is the [Pandas user guide](https://pandas.pydata.org/docs/user_guide/groupby.html).

I think this is evidenced by
 
1.  looking at common issues that are started
2. My experience getting up to speed with dedupe

For instance, looking at issues, some patterns that are common are

- How to deal with big datasets / memory management
- Using dedupe with a stream of incoming data
- What are the implications of disabling index predicates
- How long certain steps are expected to take, and what are the factors that lead to this (e.g. using sampling, number of records, num fields)
- Implementing custom Variables or comparators

In addition, when I read the API docs, often I come across options that I don't fully understand, and I wish I did.

For instance for `Dedupe.prepare_training()`, what effect does `blocked_proportion` have? Does it change precision or recall or something? Does it make it take way longer? Similarly, how should I pick `sample_size`? Or, in `Dedupe.train()` it's awesome how it explains how index_predicates affects memory, but will skipping them lower recall?

Anyway, I'd envision this as just one other page along with the rest of the docs. I can help with writing this, but on some of these points I don't have enough knowledge of the internals, so you would probably have to help. Not sure, perhaps adding this would reduce the number of issues created asking for help, but more likely it is positive net work for you :) I wanted to float the idea before I made any PRs.

NickCrews:
Or, alternatively, just add a bit more details to the API notes as I suggested in #949

fgregg:
i think this is a fantastic idea, @NickCrews. if you want to take the lead, I'd love to support you.

NickCrews:
I'll try to get to this sometime :)

PR #949, opened 2022-02-01
Tweak docs a bit more

NickCrews:
See https://github.com/dedupeio/dedupe/issues/948

### DIFF
--- a/dedupe/api.py
+++ b/dedupe/api.py
@@ -1083,7 +1083,10 @@ class ActiveMatching(Matching):
             index_predicates: Should dedupe consider predicates
                               that rely upon indexing the
                               data. Index predicates can be slower
-                              and take substantial memory.
+                              and take substantial memory. Without
+                              index predicates, you may get lower
+                              recall when true-dupes are not blocked
+                              together.
 
         """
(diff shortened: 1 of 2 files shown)


NickCrews:
Not sure why that CI failed, looks external to this patch.

Weirdly, got the same error trying to import numpy in https://github.com/dedupeio/dedupe/runs/4970736852?check_suite_focus=true, but that CI run succeeded?

NickCrews:
Looking at the [logs of a recent successful build](https://github.com/dedupeio/dedupe/runs/4970736852?check_suite_focus=true), looks like it's only a problem for python 3.10. In that run, CI tried to build fastcluster 1.2.4, failed, then kept on downgrading versions of fastcluster until it finally succeeded with fastcluster 1.1.25.

Here is [the diff between 1.1.25 -> 1.1.26](https://github.com/dmuellner/fastcluster/compare/v1.1.25...v1.1.26), not sure yet what the actual problem is...

NickCrews:
Moved that CI problem into its own issue

fgregg:
thanks!

PR #959, opened 2022-02-08
Add User Guide

NickCrews:
Fixes https://github.com/dedupeio/dedupe/issues/948

I tweaked a few other things besides the user guide, but not super attached to them, so if you want we can ditch them.


docs/User-guide.rst
```diff
@@ -0,0 +1,73 @@
...
+a look at some of the :doc:`Examples`,
+you probably want to figure out how to adapt dedupe to your own problem. This
+user guide is for that.
+
+Memory Considerations
+=====================
+
+The most common memory bottleneck is during clustering. Within every block,
+after computing a score for each pair of records, dedupe then runs a
+connected-components algorithm as a pre-processing step for the heirarchical
```
sedot42, line 15: typo: hierarchical

sedot42 (changes requested):

docs/User-guide.rst
NickCrews: Thanks! Fixed.

docs/Examples.rst
```diff
@@ -0,0 +1,17 @@
+========
+Examples
+========
+
+Dedupe is a library and not a stand-alone command line tool. To
+demonstrate its usage, we have come up with a `few example recipes for
+different sized datasets for you
+<https://github.com/dedupeio/dedupe-examples/archive/0.5.zip>`__
+(`repo <https://github.com/dedupeio/dedupe-examples>`__, as well as
```
fgregg, line 9: the "repo" is rendering strangely, it may no be doing what you intend?

fgregg (changes requested):
looking very cool overall. one change request

docs/Examples.rst
fgregg: Screenshot 2022-02-09 at 15-07-22 Examples — dedupe 2 0 11 documentation

fgregg:
thanks @NickCrews , i’ll review more closely

docs/User-guide.rst
```diff
@@ -0,0 +1,73 @@
+**********
+User Guide
+**********
+
+Once you understand :ref:`how dedupe works <how-it-works-label>`, and you've taken
```
fgregg, line 5: i think this paragraph might need a different framing? the current sections are all about what to do if you have really big data or you are not getting accurate results. Maybe this needs more of a troubleshooting frame and not a "how to use" framing?

docs/User-guide.rst
```diff
@@ -0,0 +1,73 @@
...
+user guide is for that.
+
+Memory Considerations
+=====================
+
+The most common memory bottleneck is during clustering. Within every block,
+after computing a score for each pair of records, dedupe then runs a
+connected-components algorithm as a pre-processing step for the hierarchical
+agglomerative clustering.
+`Currently, this requires having the entire edge list in memory
```
fgregg, line 17: this is not right, the edgelist stays as memmapped array, and mainly on disk. we do need to store what are basically pointers (array indices) to edges in memory but that can be a lot smaller than the edgelist (this is the components dict in the union_find algorithm).  

still the compent dict can take a lot of memory

docs/User-guide.rst
```diff
@@ -0,0 +1,73 @@
...
+<https://github.com/dedupeio/dedupe/issues/819>`__, which for :math:`N`` records
+leads to :math:`O(N^2)` memory usage. If you end up with any "superblocks" because
+your blocking rules weren't picky enough, this can run you out of memory. You
```
fgregg, line 20: super blocks won't really create a memory problem. they could make things too slow, but we filter the components to not be bigger than 30,000 vertices

docs/User-guide.rst
```diff
@@ -0,0 +1,73 @@
...
+
+The other major memory bottleneck is building the index for blocking. If this is
+a problem, you can try turning off index blocking rules (and just use predicate
+blocking rules) by setting ``index_predicates=False`` in
+:meth:`dedupe.Dedupe.train`.
+
+Time Considerations
+===================
+
+Similar to the memory considerations, the slow part of dedupe is whenever
```
fgregg, line 31: i think clustering was a major bottleneck until https://github.com/dedupeio/dedupe/pull/912, i doubt it's where the problem is now.

i would expect that blocking can be pretty slow depending on how many predicates you learned and what type.

docs/User-guide.rst
```diff
@@ -0,0 +1,73 @@
...
+we have to run an algorithm that has a time complexity :math:`> O(N)`. The
+scoring of pairs ():math:`O(N^2)`) is already implemented in a parallelized way,
+but the clustering is not (yet?), so again this clustering step is probably
+going to be your bottleneck. Again, inspecting and splitting up superblocks
+can help.
+
+
+Improving Accuracy
```
fgregg, line 39: this seems like a good start!

fgregg (changes requested):

docs/User-guide.rst
NickCrews: Good idea, will do

docs/User-guide.rst
NickCrews: OK, makes sense. Sorry, I was just reading past issues and PRs and this was my best summary, but admittedly I wasn't confident in my understanding.

What do you think *are* the main memory bottlenecks then? Someone comes here after their machine runs out of memory, what are the first things to check or try? Thanks!

docs/User-guide.rst
NickCrews: OK I can just remove this

docs/User-guide.rst
NickCrews: OK I can swap it to this. In that case, the advice for dealing with slow blocking would be:
- Turn off index predicates
- inspect what predicates you learned. If a lot (>20??) of them, then you should...
  - Remove some of your training samples?
  - Reduce `recall` in your call to `train()`?

Is that correct?

docs/User-guide.rst
NickCrews: Thanks!

NickCrews (commented):
Thanks for review! Can you answer those couple questions and then I'll submit a new draft?

I can stack new commits on top instead of force-pushing and overwriting so that we have the revision history. And then once finalized I can squash it all back together.

docs/User-guide.rst
fgregg: i think the index predicates and the components dicts are probably the most likely places, in that order.

docs/User-guide.rst
fgregg: yeah, i would recommend reducing the number of positive examples in the the training data and turn off index predicates. i would make sure that people understand they are going to be trading off recall for speed probably.

this is also an area where can probably benefit from parallelization or other code strategies, although current attempts haven't really proved promising yet.

fgregg:
thanks @NickCrews, responded!

NickCrews:
Thanks! I'll get to this on Sunday or Monday probably, I'm off the next few days.

NickCrews:
Darn, I was worried that force push was gonna delete our conversation. It did. Hopefully you can still see what's happening

fgregg:
thanks @NickCrews !

NickCrews:
Thanks for the reviews and the quick merge @fgregg !

PR #961, opened 2022-02-10
Pin docs requirements

NickCrews:
This is following the example of
https://github.com/datalad/datalad/pull/6346

The bullet points on the online docs aren't rendering correctly.


NickCrews:
Currently docs look like:

image

With this PR, docs look like:

image

PR #962, opened 2022-02-10
Tweaks

NickCrews:
Few trivial things


.github/workflows/pythonpackage.yml
```diff
@@ -77,6 +76,7 @@ jobs:
         pip install build
         python -m build --sdist
     - name: Publish wheels to PyPI
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
```
fgregg, line 79: we don't really need to build the wheels unless we are going to push them to pypi, so let's keep this as is.

fgregg (changes requested):
docs change looks good. i want to keep the CI conditional as is.

Thanks for the suggestions!

.github/workflows/pythonpackage.yml
NickCrews: It's possible that some PR could break how wheels are built, and we wouldn't actually know that until you cut a new release and trigger this workflow. Admittedly not super likely. With this change, you would see the bug earlier. Still want to cut it out?

fgregg: building the wheels adds 20 minutes, and i'd like normal tests to be fast, so yes.

it might be nice to run the wheelbuilding once a day (or once a day if there was a change a la https://github.community/t/trigger-action-on-schedule-only-if-there-are-changes-to-the-branch/17887/4), but that would be a separate PR

NickCrews: Roger, sounds good I'll drop that commit :)

Issue #968, opened 2022-02-21
weird formatting on docs


fgregg:
@NickCrews , do you know why there's this double line?

![Screenshot 2022-02-21 at 13-29-05 Library Documentation — dedupe 2 0 11 documentation](https://user-images.githubusercontent.com/536941/155009883-2506faea-2f1c-4f87-ad88-aa546899cc62.png)

PR #969, opened 2022-02-21
Fix docs CSS (weird formatting on docs #968)

NickCrews:
Fixes https://github.com/dedupeio/dedupe/issues/968.

Don't really know what this custom CSS was trying to do, but it looks better now, hopefully it didn't make something else look worse that I didn't find.

But, I'm of the opinion that all else equal, less CSS is better

Before:
image

After:
image

Before:
image

After:
image


fgregg:
thanks!

Issue #991, opened 2022-04-25
shoud we integrate with scikit learn


fgregg:
Thinking about #990, i'm wondering if we should make scikit learn a dependency.

When we started dedupe 10 years ago, it was really hard to get scipy and scikit learn set up on users machines, but it's not anymore.

It would be nice to get out of the game of implementing some algos ourselves.

Thoughts, @fjsj @NickCrews

fjsj:
I agree, IMO scikit-learn is fairly default and sometimes I see it being used just for model evaluation / visualization, so it ends up being a very common dependency in ML projects.

NickCrews:
I think this move would be well worth it.

Presumably after doing this it would be easy to swap out different classifiers, so a to-do is figuring out how to configure which classifier you want to use.

fgregg:
this is actually very easy, because i've already made so the classifier has the sklearn api.

fjsj:
Here I adapted the code via inheritance to support sklearn classifiers: https://github.com/vintasoftware/deduplication-slides/blob/pycon-2020/rf_dedupe.py

But it's worth checking the conditionals on `def fit` against the current Dedupe version.

fgregg:
closed by #992

NickCrews:
Nice job @fgregg, this looks awesome!

Issue #990, opened 2022-04-25
Consider making the default classifier a random forest


fgregg:
I bet it would generally be better at picking up some of the funny relationships.

NickCrews:
I think this is a very good motivation for #991

fgregg:
I ended up not switching to random forest in #992 and continue to use regularized logistic regression. 

the loss in precision for the canonical case was a bit more than i was fully comfortable with, and wanted more time to think about.

not quite ready to close though.

Issue #967, opened 2022-02-21
Add benchmarks


NickCrews:
Branched off of https://github.com/dedupeio/dedupe/issues/965#issuecomment-1046319666.

EDIT: See next comment for using ASV instead of @profile

Place @profile decorators on bottleneck functions using `memory_profiler`

List this dependency as extra, so that most users don't need to install it.

Also, to prevent overhead from @profile getting run always (even when we don't want profiling), wrap it in our own custom decorator that is usually a noop:

```python
def dd_profile(func, *args, **kwargs):
    # Maybe a better way to configure this? Would have to be at import time
    if os.environ["DEDUPE_PROFILE"]:
        # Actually add the profiler wrapper
        return profile(func, *args, **kwargs)
    else:
        # noop
        return func
```

Next steps are to probably actually make a new branch and apply it to some of the examples?

NickCrews:
OK, I think a better option is [Air Speed Velocity](https://asv.readthedocs.io/en/stable/index.html).

- Don't have to modify the codebase at all, benchmarks are separate, like tests
- measure time, memory, and custom attributes (eg we could measure accuracy)
- used by big names like numpy, pandas, scipy
- See examples at that link of static websites, it has built in support for github pages
- Could run on every PR, or could trigger only when we care from a comment like `@benchmark main..HEAD --bench benchmarks.TimeSuite.time_range` [as pandas does](https://github.com/pandas-dev/pandas/blob/1951dd70dee5df23cc85e09d44f745bc3e724dcd/.github/workflows/asv-bot.yml)

Initially, I would think that we would get a lot of value out of measuring
- time
- peak memory
- accuracy
on just one or two end-to-end runs (training and evaluating separate). Either could use [the test datasets](https://github.com/dedupeio/dedupe/tree/master/tests/datasets), or could use the examples from [dedupe examples](https://github.com/dedupeio/dedupe-examples)

One tricky thing would be to decide how to store [the results](https://asv.readthedocs.io/en/stable/using.html#benchmarking) (eg the metrics collected from past runs). Would want a way to ensure a consistent machine (using github actions? and/or with docker?). Keep the results in this repo, or keep them in a separate repo, possibly using submodules as they mention?

@fgregg @fjsj any thoughts here?

fgregg:
I love this ides @NickCrews . if you want to try to get something rough and ready set up, i think that would a really good step forward. if it's looking very valuable, then we will figure out the tricky bits.

[...]

PR #992, opened 2022-04-26
start to use sklearn for ml algorithms

fgregg:
relates to #991 and #990 

## Todo 
- [x] get test passing
- [ ] <s>replace `haversine` dependency with https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html#sklearn.metrics.pairwise.haversine_distances</s>
- [ ] <s>replace `simple-cosine` dependency with https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.cosine_similarity.html#sklearn.metrics.pairwise.cosine_similarity</s>

From [trying to replace the cosine distance](https://github.com/dedupeio/dedupe/pull/992/commits/9bc5d8b3645f87672aeb07563a53d9d91a2935ba), it's pretty clearly not worth doing that unless we can batch the calls. 

Any other likely places where we can use sklearn or scipy code instead of an additional library or replace dedupe code, @fjsj , @NickCrews ?


fjsj:
@fgregg 
Another thing you could replace is [the TF-IDF index](https://github.com/dedupeio/dedupe/blob/master/dedupe/canopy_index.py#L14), although it's not trivial to build scikit's TF-IDF iteratively and query it like an index, if I recall correctly.

I also see some nested `for`s in [canonical.py](https://github.com/dedupeio/dedupe/blob/master/dedupe/canonical.py), but I'm not sure if something in scikit-learn or scipy can replace that.

There are also graph operations at [clustering.py](https://github.com/dedupeio/dedupe/blob/master/dedupe/clustering.py) that could in principle use sparse graphs from scipy, but it would be a full rewrite and I'm not sure if performance / memory usage would improve / reduce.

dedupe/api.py
```diff
@@ -987,7 +987,7 @@ class ActiveMatching(Matching):
     """
     Class for training a matcher.
     """
-    classifier = rlr.RegularizedLogisticRegression()
+    classifier = sklearn.ensemble.RandomForestClassifier()
```
NickCrews, line 990: IDK why someone might have mutliple ActiveMatching instances at the same time, but if so they would share the same classifier. Ideally this should be an instance attribute

NickCrews:
I don't think I'm familiar enough with the code to be that confident, but from a cursory scan it seems like you got the obvious low-hanging fruit.

NickCrews:
If we implemented #967 then perhaps we could do a whole lot less guessing about how performance would be changed

dedupe/api.py
fgregg: that's a good call, but an API change. I'll open an issue for that and do that in another PR.

fgregg:
nice, so @NickCrews's benchmarking set up let's us see that sklearn's Random Forest classifier leads us to use about 90M of peak memory as opposed to 50M of peak memory.

fgregg:
@NickCrews,there seems to be something going on with the benchmarks. The precision and recall scores are exactly the same between this PR and the comparison branch. That does not seem possible.

fgregg:
I adjusted the benchmarker, and we are getting a much clearer picture.

* Recall is up. Almost always at 1. 
* For record linkage modes, precision stays high, sometimes small improvement
* For deduplication mode, recall is up to 1, but the precision has dropped from low 90s to high 70s.

NickCrews:
I think the issue might be
- benchmark runs, settings file created
- switch to new commit, settings file is in .gitignore so it stays unchanged
- on new commit, the benchmark still uses the old settings file

fgregg:
> * Recall is up. Almost always at 1.
> 
>     * For record linkage modes, precision stays high, sometimes small improvement
> 
>     * For deduplication mode, recall is up to 1, but the precision has dropped from low 90s to high 70s.

If we increase the threshold a bit for canonical, we can get better recall and precision than with logistic regression. I think I'll keep this change.

The only thing that we have left to do is make sure that existing settings file that use rlr can still be loaded, and add a deprecation notice if we find such a settings file.

fgregg:
@github-actions benchmark

fgregg:
@benchmark

fgregg:
@NickCrews this is cool as hell. isn't *quite* there yet, as it uses the same commit for before and after.

NickCrews:
Lol DAMMIT

fgregg:
thanks for your work on this @NickCrews ! this is really nice!

NickCrews:
benchmark bot fixed in https://github.com/dedupeio/dedupe/pull/1018

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark canonical.Canonical.track_precision

fgregg:
@benchmark -b canonical.Canonical.track_precision

fgregg:
@benchmark

fgregg:
@benchmark -b canonical.Canonical.track_precision

fgregg:
@benchmark -b canonical.Canonical.track_precision

fgregg:
@benchmark -b canonical.Canonical.track_precision

fgregg:
@benchmark

fgregg:
@benchmark -b canonical.Canonical.time_run

fgregg:
@benchmark -b canonical.Canonical.peak_mem

fgregg:
@benchmark -b canonical.Canonical.peakmem_run

fgregg:
@benchmark -b canonical.Canonical.track_precision

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@NickCrews it lives!

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

fgregg:
@benchmark

Issue #984, opened 2022-03-21
treat "unsure" as yes


fgregg:

NickCrews:
Can you elaborate a little more? Just to make the code for console labeling simpler? Or does this actually improve matching performance (better to err on the side of it being a match)? At this point, in my use, when I'm labeling, should I just always say "yes" when I'm tempted to say "unsure"?

fgregg:
I saw a presentation on record linkage that compared human reviewers and a ground-truth mechanism, and the main difference in the the ground-truth performance was that humans are much to conservative in accepting matches.

NickCrews:
Awesome, thanks! Think the end goal should be to remove "unsure" from the UI, or should keep it in the UI and just treat it as "yes" under the hood. The second is a bit disingenuous, but per what you just said, it might actually lead to better performance in actual use. Could mention a note that says "If you're unsure, just say yes", but that might not be seen.

fgregg:
I'm thinking of the second option (lying to people), with a flag that people can set if they really don't want that behavior. The more sophsticated thing would be to give it a "half label" or, equivalently add the record as both a positive and as a negative.

fgregg:
Actually, i like the idea of adding it as both a positive and negative.

NickCrews:
Adding as both seems reasonable, what are the implications of that?
- Still will learn a blocking rule for it (good)
- Won't affect weights of the learner as they will cancel each other out (good)
- ???

fgregg:
if ever introduce sample weights, it could be good to give each of the duplicated records a half weight. i think this is a solid idea to try out!

NickCrews:
> if ever introduce sample weights, it could be good to give each of the duplicated records a half weight.

hmmm I see what you mean. Does bring up the question of what is "half weight". I would assume that it would be an option to assign arbitrary weights to samples (ie range [0, inf]), so if you said unsure, does that mean just divide the initial weight in half?

Did I miss any other implications that are important?

> in the meantime, i think this is a solid idea to try out!

Cool, this seems like it would be really easy to implement, just change the behavior in [console_label()](https://github.com/dedupeio/dedupe/blob/c2ac55932930591ccfdbce650f66715a52425c36/dedupe/convenience.py#L179), right? Nowhere else to worry about?

fgregg:
> does that mean just divide the initial weight in half?

i think so, yes.



> Did I miss any other implications that are important?

[...]

PR #994, opened 2022-04-27
Treat "unsure" as both "yes" and "no"

NickCrews:
closes https://github.com/dedupeio/dedupe/issues/984

### DIFF
--- a/dedupe/convenience.py
+++ b/dedupe/convenience.py
@@ -92,13 +92,19 @@ def _print(*args) -> None:
     print(*args, file=sys.stderr)
 
 
-LabeledPair = Tuple[TrainingExample, Literal["match", "distinct"]]
+LabeledPair = Tuple[TrainingExample, Literal["match", "distinct", "unsure"]]
 
 
 def _mark_pair(deduper: dedupe.api.ActiveMatching, labeled_pair: LabeledPair) -> None:
     record_pair, label = labeled_pair
     examples: TrainingData = {"distinct": [], "match": []}
-    examples[label].append(record_pair)
+    if label == "unsure":
+        # See https://github.com/dedupeio/dedupe/issues/984 for reasoning
+        examples["match"].append(record_pair)
+        examples["distinct"].append(record_pair)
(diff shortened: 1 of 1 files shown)


dedupe/convenience.py
```diff
@@ -88,101 +88,101 @@ def randomPairsWithReplacement(n_records: int, sample_size: int) -> IndicesItera
...
     buffer_len = 1  # Max number of previous operations
-    examples_buffer: List[Tuple[TrainingExample, Literal['match', 'distinct', 'uncertain']]] = []
-    uncertain_pairs: List[TrainingExample] = []
+    unlabeled: List[TrainingExample] = []
+    labeled: List[LabeledPair] = []
 
     while not finished:
         if use_previous:
-            record_pair, _ = examples_buffer.pop(0)
+            record_pair, _ = labeled.pop(0)
```
fgregg, line 127: we would need to pop twice in the case of unsure, right?

fgregg (changes requested):

dedupe/convenience.py
```diff
@@ -88,101 +88,101 @@ def randomPairsWithReplacement(n_records: int, sample_size: int) -> IndicesItera
...
-                examples: TrainingData
-                examples = {'distinct': [],
-                            'match': []}
-                examples[label].append(record_pair)
-                deduper.mark_pairs(examples)
+            unlabeled.append(record_pair)
 
-    for record_pair, label in examples_buffer:
-        if label in ['distinct', 'match']:
+        while len(labeled) > buffer_len:
```
fgregg, line 181: we'll also need to do this a bit differently, since unsure adds to elements to labeled

fgregg (changes requested):

dedupe/convenience.py
NickCrews: Oops, yeah. See fixup commit

fgregg (approved):

fgregg:
thanks, @NickCrews !

Issue #996, opened 2022-04-29
Accept training data directly, not just in a filelike


NickCrews:
Hey, so this is a bit of a resurrection of https://github.com/dedupeio/dedupe/issues/935, so give the same answer now if you want. I agree it's not a huge value add to support file paths, but I have a slightly different need.

I have some `TrainingData` (ie dict with keys "match" and "distinct", values are lists of record pairs) that I constructed from a pandas dataframe (using `dedupe.training_data_dedupe` no less). Right now, I can pass that to `Dedupe.prepare_training` as `StringIO(json.dumps(my_data))`, but then Deduper just deserializes it again.

I propose that `prepare_training` should accept a dict as well (also maybe a json string? but that is a weird in between that I don't think a user would encounter). It's funny that dedupe has a convenience method `training_data_dedupe` to construct a `TrainingData`, but then you have no way to use it directly.

Don't worry about the other IO methods that I mentioned in the first issue, because those all deal with opaque binary data that the user will never construct directly.

NickCrews:
Err, wait, I can just pass it to `mark_pairs`

I think it's a bit confusing how there are multiple ways to pass in training data (both in prepare_training and in mark_pairs). Ideally I think prepare_training should just rely upon training data that was passed in previously. eg

```python
data: Data = load_data()
td: TrainingData = dedupe.load_training(filelike)
deduper.mark_pairs(td)
deduper.prepare_training(data)
# at this point why do we even have separate prepare_training and train steps?
# Everything could just happen in train()?
deduper.train()
```

Of course this is a breaking API change

fgregg:
i agree that the flow is not really optimized for people who already have training data and who do not want to add more training data.

i'm not sure how common a use case this.

fgregg:
the key thing that prepare_trainign does for us is to enable uncertain_pairs to run.

NickCrews:
> the key thing that prepare_trainign does for us is to enable uncertain_pairs to run.

Hmm yeah that makes sense. So we still need separate prepare_training() and train() methods for the purpose of active labeling.

OK, well could still remove the training_file argument to prepare_training, and just have mark_pairs be responsible for all labeling. Or, to be non-breaking, and to fix my need here, just unify the code paths so that mark_pairs can take either a file-like or a dict, and then prepare_training, if given training data, just hands it off to mark_pairs.

NickCrews:
[...]

PR #997, opened 2022-04-29
Add read_training and write_training to API

NickCrews:
Fixes https://github.com/dedupeio/dedupe/issues/996


NickCrews:
Maybe need to add a test?

PR #998, opened 2022-05-01
REF: remove unneeded ActiveMatching._sample()

NickCrews:
No behavior changes (I don't think)

As far as I could tell, this private method is only used in
prepare_training

- Had default arguments that didn't match prepare_training
- Had stale docstring
- Now verifies data before resetting active learner, as a nicety
- Fixes message from assert if you
don't call prepare_training()


fgregg:
beautiful catch! thank you!

NickCrews:
I hate code and I want as little of it as possible :)

PR #1003, opened 2022-05-03
Format with Black

### LINKED ISSUE #1001: FEAT: Use black formatting
Feature request for developing, not a huge deal but I find it very helpful.

I love the automatic formatting that [black](https://github.com/psf/black) does. It's becoming quite a standard for python projects. Do you want to change to that? formatting all existing code would add some noise to git blame, although [that can be mitigated](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html). Besides that all that would need to happen is add to requirements and then add a CI check alongside the other static linters.

NickCrews:
closes #1001

After this lands and we know the commit hash, we can create a `.git-ignore-revs` file


fgregg:
thanks @NickCrews

PR #1030, opened 2022-06-01
fix bug caused by deleting keys as we iterated through dict

### LINKED ISSUE #1029: CanopyIndex stop word removal non-determinism
I trained a dedupe model for address matching and ended up seeing different partition() results for the same input across different executions. I traced the non-deterministic behavior to one of my predicates (`TfidfNGramCanopyPredicate: (0.6, address1)` as seen in some log output. More specifically, I would see different sets of stop words being removed across different runs by looking at this log [line](https://github.com/dedupeio/dedupe/blob/main/dedupe/canopy_index.py#L30). 

I put in some debug logs and noticed that the tfidf counts seemed to always be the same but the stop words removed were not always the same which ended up correlating to different end results. I added a log message before this [line](https://github.com/dedupeio/dedupe/blob/main/dedupe/canopy_index.py#L28) to further check out what was happening.

```python
       for wid, docs in self.index._wordinfo.items():
            logger.info(f"  {wid} {self.lexicon._words[wid]} {len(docs)}")   # new log
            if len(docs) > threshold:
                word = self.lexicon._words[wid]
                logger.info("Removing stop word {}".format(word))
                del self.index._wordinfo[wid]
                continue
```

One example set of log output is this:

```
INFO:dedupe.canopy_index:  1  F 277
INFO:dedupe.canopy_index:  2  S 3041
INFO:dedupe.canopy_index:Removing stop word  S
INFO:dedupe.canopy_index:  4 34 252
INFO:dedupe.canopy_index:  5 4  805
```

I noticed the the wid of 3 was missing which tells me it was skipped. I then changed the code to be like this and re-ran:

```python
       for wid, docs in self.index._wordinfo.items():
            logger.info(f"  {wid} {self.lexicon._words[wid]} {len(docs)}")   # new log
            if len(docs) > threshold:
                word = self.lexicon._words[wid]
                logger.info("Removing stop word {}".format(word))
                #del self.index._wordinfo[wid]
                #continue
```

Then I would see output like this (note that it seems like the iteration order is not defined but I think that is expected):

```
INFO:dedupe.canopy_index:  1  D 1064
INFO:dedupe.canopy_index:Removing stop word  D
INFO:dedupe.canopy_index:  2  S 3041
INFO:dedupe.canopy_index:Removing stop word  S
INFO:dedupe.canopy_index:  3 12 561
INFO:dedupe.canopy_index:  4 26 266
INFO:dedupe.canopy_index:  5 6  737
```

In this case, by not calling `del` on the wordinfo entry, it did not skip the next entry after removing a stop word. So it looks like the non-determinism I was seeing in my end partition output is due to different stop words being removed due to this iteration bug across different executions. This situation of different stop words being removed will occur when they happen to be next to each other in wid order. I haven't yet been able to trace through the code to figure out what exactly is going wrong in the `del` operation. Maybe this is incorrect usage or there is some internal bug in the data structure but not sure. 

I am using python 3.8.10 on MacOS with dedupe-2.0.13

fgregg:
closes #1029 

@oreccb can you check out this branch and confirm it fixes your problem?

### DIFF
--- a/dedupe/canopy_index.py
+++ b/dedupe/canopy_index.py
@@ -21,15 +21,18 @@ class CanopyIndex(TextIndex):  # pragma: no cover
         N = len(self.index._docweight)
         threshold = int(max(1000, N * 0.05))
 
-        self._wids_dict = {}
+        stop_words = [
+            wid for wid, docs in self.index._wordinfo.items() if len(docs) > threshold
+        ]
+        for wid in stop_words:
+            word = self.lexicon._words[wid]
+            logger.info("Removing stop word {}".format(word))
+            del self.lexicon._words[wid]
+            del self.index._wordinfo[wid]
 
         bucket = self.index.family.IF.Bucket
+        self._wids_dict = {}
         for wid, docs in self.index._wordinfo.items():
-            if len(docs) > threshold:
-                word = self.lexicon._words[wid]
-                logger.info("Removing stop word {}".format(word))
-                del self.index._wordinfo[wid]
-                continue
             if isinstance(docs, dict):
                 docs = bucket(docs)
             idf = numpy.log1p(N / len(docs))
--- a/dedupe/canopy_index.py
+++ b/dedupe/canopy_index.py
@@ -37,9 +37,8 @@ class CanopyIndex(TextIndex):  # pragma: no cover
             self._wids_dict[term] = (wid, idf)
 
         for wid in stop_words:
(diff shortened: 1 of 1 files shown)


oreccb:
@fgregg I added in this change locally and can confirm that it fixes the issue!

fgregg:
@benchmark

PR #1034, opened 2022-06-04
Datamodel-refactor

NickCrews:
@fgregg ready for a review whenever, thanks! I can split into smaller PRs if easier, but I think individual commits are manageable if you review those one at a time.

This is a refactor, no real change besides the one in `Move all_subclasses() to classmethod`.

- add type hints
- use new type hint notation
- the biggest thing is updating variable name usage in datamodel.py to be consistent with docs usage. Keeping field vs variable vs "variable definition" all consistent.
- couple of other small tweaks in the internals of datamodel, eg using generators to be a little cleaner, making __len__ calculated more cleanly, removing a few temp variables.
- don't fail-fast on tests

### DIFF
--- a/dedupe/variables/base.py
+++ b/dedupe/variables/base.py
@@ -1,14 +1,14 @@
 from __future__ import annotations
 
-from typing import Callable, Sequence, Type, Any, Iterable
+from typing import Callable, ClassVar, Sequence, Type, Any, Iterable
 
 from dedupe import predicates
 
 
 class Variable(object):
     name: str
-    type: str
-    predicates: Sequence[Callable[[Any], Any]]
+    type: ClassVar[str]
+    predicates: list[Callable[[Any], Any]]
 
     def __len__(self):
         return 1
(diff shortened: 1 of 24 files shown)


fgregg:
@NickCrews, this is looking cool. I marked it as a WIP, but if this ready, just add a description and I'll review.

NickCrews:
@fgregg just added a description, ready to review :)

dedupe/variables/base.py
```diff
@@ -1,9 +1,15 @@
...
+
+from typing import Callable, ClassVar, Sequence, Type, Any, Iterable
 
 from dedupe import predicates
 
 
 class Variable(object):
+    name: str
+    type: ClassVar[str]
+    predicates: list[Callable[[Any], Any]]
```
fgregg, line 11: a predicate has to return something iterable.

fgregg (approved):
looks good, Nick. I think we can tighten the typing of predicates, but that's optional.

NickCrews:
@fgregg I agree with all the comments you added, nice job adding them in those later followup commits. Thanks!

PR #1040, opened 2022-06-06
More typing

fgregg:
This PR adds typing to the rest of the library and narrows a typing in a few places.


fgregg:
thus us about done. just got to get it working on earlier versions of python

Issue #1043, opened 2022-06-06
factor linters out into separate step


fgregg:
we only need to these once.

-flake8 
-black --check
-mypy
-isort --check

maybe factor these all into single step and then fan out to run the test and integration tests on all the machines.

PR #1055, opened 2022-06-09
Do CI linting on only one machine

NickCrews:
Fixes https://github.com/dedupeio/dedupe/issues/1043

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -3,6 +3,31 @@ name: tests
 on: [push, pull_request]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -3,6 +3,31 @@ name: tests
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: |
+          pip install --upgrade pip
+          pip install -e .
```
fgregg, line 17: i don't think we actually need to install dedupe.

fgregg (changes requested):

.github/workflows/pythonpackage.yml
NickCrews: Indeed, fixed

PR #1064, opened 2022-06-12
import numpy as np

NickCrews:
Total nit, but it is more constent with how others use it, and it's a bit less verbose.


fgregg:
sorry, not going to do this one.

Issue #1052, opened 2022-06-09
Consider configuring "index_predicates" flag in variable definition


NickCrews:

fgregg:
i like this idea. some considerations:

1. what's the default? i think it makes sense that you have to toggle on index predicates
2. not all variable types have index predicates, we would need to provide some reasonable warning when that flag is a no-op
3. what does the transition story look like. there are a few cases
   1.  in `train`, `index_predicates=False`: disable all index predicates regardless of what the variable definition is 
   2. in `train`, `index_predicates=True` do whatever the variable definitions say.
   3. in `train`, `index_predicates` is not explicitly set. If there are variable definitions that ask for index predicates, then i think we should do what the variable definitions ask
   4. in `train`, `index_predicates` is not explicitly set. if there are no variables that ask for index_predicates (current situation), then probably raise an warning?
4 would be want to preserve a flag to disable all index_predicates in the train method... maybe.

NickCrews:
> what's the default? i think it makes sense that you have to toggle on index predicates

I agree, by default don't use index predicates

> not all variable types have index predicates, we would need to provide some reasonable warning when that flag is a no-op

I would throw an error. Plenty of other variables have params and settings that are specific to them, I view this change as lumping this option in with the other ones.

> would be want to preserve a flag to disable all index_predicates in the train method... maybe.

I think the end goal should be to remove the `index_predicates` argument from `train()`. The issue as you say is the transition period. I would think that should keep past behavior:

```python
    def train(
        self, recall: float = 1.00, **kwargs
    ) -> None:  # pragma: no cover
        index_predicates = kwargs.get("index_predicates", None)
        if index_predicates is not None:
            warnings.warn("the arg deprecated")
            if index_predicates:
                # override variable config, pretend "index":True was used in all definitions
            else:
                # override variable config, pretend "index":False was used in all definitions
        # continue onwards
  
```
After some time, we remove the **kwargs and this logic.

This agrees with your suggestions `i` and `iii`, but disagrees with `ii`. If we follow `ii` and have variable indexes off by default, then this is going to change behavior for people who don't update their code. I don't think I understand the situation you are talking about for `iv`?

PR #1065, opened 2022-06-12
Refactor labeler.py

NickCrews:
See each commit individually, nothing functional changes

EDIT: still shouldn't be any funcitonal changes, though this has gotten much more substantial


dedupe/labeler.py
```diff
@@ -69,24 +69,22 @@ def candidates(self) -> TrainingExamples:  # type: ignore[override]
...
-        self.distances = self.transform(self._candidates)
+        self.distances = self._distances(self._candidates)
 
         random_pair = random.choice(self._candidates)
         exact_match = (random_pair[0], random_pair[0])
-        self.fit_transform([exact_match, random_pair], [1, 0])
+        self.fit([exact_match, random_pair], [1, 0])
 
-    def transform(self, pairs: TrainingExamples) -> numpy.typing.NDArray[numpy.float_]:
+    def _distances(self, pairs: TrainingExamples) -> numpy.typing.NDArray[numpy.float_]:
```
fgregg, line 78: nice.

dedupe/labeler.py
```diff
@@ -69,24 +69,22 @@ def candidates(self) -> TrainingExamples:  # type: ignore[override]
...
         return self.data_model.distances(pairs)
 
-    def fit(self, X: numpy.typing.NDArray[numpy.float_], y: LabelsLike) -> None:
-
+    def _fit(self, X: numpy.typing.NDArray[numpy.float_], y: LabelsLike) -> None:
         self.y: numpy.typing.NDArray[numpy.int_] = numpy.array(y)
         self.X = X
```
fgregg, line 83: i know this was nor your design, but it's a bit weird to have the fit method be responsible for setting these instance attributes. `mark` would be a more natural place, imo

dedupe/labeler.py
```diff
@@ -69,24 +69,22 @@ def candidates(self) -> TrainingExamples:  # type: ignore[override]
...
+        sklearn.linear_model.LogisticRegression.fit(self, self.X, self.y)
```
fgregg, line 84: i think it's good to have an explicit call here, but if we do that, we shouldn't also have the class inherit from `sklearn.linear_model.LogisticRegression`

dedupe/labeler.py
```diff
@@ -390,12 +383,11 @@ def mark(self, pairs: TrainingExamples, y: LabelsLike) -> None:
...
 
-        self.fit_transform(self.pairs, self.y)
-
-    def fit_transform(self, pairs: TrainingExamples, y: LabelsLike) -> None:
+        self.fit(self.pairs, self.y)
 
+    def fit(self, pairs: TrainingExamples, y: LabelsLike) -> None:
         for learner in self.learners:
-            learner.fit_transform(pairs, y)
+            learner.fit(pairs, y)
```
fgregg, line 390: `fit` is a bad verb for something that both sets the data on a learner and retrains the learner. I think maybe we want to get rid of fit from the public API and have `mark` be responsible for all of this.

fgregg (changes requested):

dedupe/labeler.py
NickCrews: If I turn this to `super().fit(self.X, self.y)` then mypy complains because it needs both parents in the inheritance to have the same signature for `fit`: `dedupe/labeler.py:84: error: Argument 1 to "fit" of "Learner" has incompatible type "ndarray[Any, dtype[floating[_64Bit]]]"; expected "List[Tuple[Mapping[str, Any], Mapping[str, Any]]]"  [arg-type]`

I think we should rename `RLRLearner` to `MatchLearner` (it is comparing records using distances, and this is sorta a nice opposition to BlockLEarner) and make it wrap `LogisticRegression` (or RandomForest, or whatever) using composition, not inheritance.

dedupe/labeler.py
NickCrews: I agree. still trying to figure out how to deal with the fact that the learners are stateful with their data. I almost want to split them up, so that one component is stateless like sklearn models and has a `fit()`, and that is owned by a learner that is stateful and remembers the old training data, and that has the `mark()`

NickCrews: OK in this new version all the sublearners are stateless (in regards to saving training data) and don't have a `mark`, just a stateless fit(). Only DisagreementLEarner has a `mark` that stores traingin data.

dedupe/labeler.py
NickCrews: See if you like what I did.

dedupe/labeler.py
NickCrews: See new version and comment above.

NickCrews:
@fgregg this has gotten way bigger, and this was very much done in an experiemental/haphazard way, so that some commits undo/modify what previous commits did. I can rebase this in a more sensible way if the large scale idea looks right. Maybe though take a look at the final result and see if this is generally heading in a good direction before I do all that cleanup work. Or I can split this into a PR with the less contentious changes, get that merged, and then take a look at the more structural changes afterwards.

dedupe/labeler.py
```diff
@@ -26,155 +26,107 @@
 logger = logging.getLogger(__name__)
 
 
-class Learner(ABC):
-    candidates: TrainingExamples
+class HasCandidates:
+    """Has a read-only list of pairs that we could ask the user to label."""
```
fgregg, line 30: candidates is not really read only since we remove records from it.

dedupe/labeler.py
```diff
@@ -26,155 +26,107 @@
...
-        return weighted_bias
+    def remove(self, index: int) -> None:
+        self._candidates.pop(index)
+        self._distances = numpy.delete(self._distances, index, axis=0)
 
     def candidate_scores(self) -> numpy.typing.NDArray[numpy.float_]:
-        scores: numpy.typing.NDArray[numpy.float_] = self.predict_proba(self.distances)[
-            :, 1
-        ].reshape(-1, 1)
+        if not self._fitted:
```
fgregg, line 74: if we are not fitted, then let's have the candidate scores return an array of "0.5" of length candidates

fgregg: or an error.

dedupe/training.py
```diff
@@ -138,9 +139,15 @@ def score(predicate: Predicate) -> float:
 
         return candidates
 
-    def cover(self, pairs: TrainingExamples) -> Cover:
+    def cover(self, pairs: TrainingExamples, index_predicates: bool = True) -> Cover:
```
fgregg, line 142: don't love this change. is it really ugly to avoid this?

fgregg (changes requested):
Really liking where this is going. A few comments in line.

fgregg:
With your refactor, it's pretty apparent that the weighted sampling methods on these classes are a bit odd. We can save that for a future PR though.

dedupe/labeler.py
NickCrews: Yeah RO is a bit misleading, it is RO in the sense that you can't modify it directly, but you can modify it via remove(). But just better to leave out the RO part, I think you're right.

dedupe/labeler.py
NickCrews: I like ValueError more. Done.

dedupe/training.py
NickCrews: Have you seen the other changes from this commit? and the corresponding commit message? I think the current way that we deal with this, where in DisagreementLearner.learn_predicates() we do `self.blocker.block_learner.blocker.predicates.copy()` is a terrible code smell, reaching 3 levels deep, where `blocker.block_learner.blocker` are all mean the same thing. So I wanted to actually create public APIs at each level.

The other problem there is the pattern of "temporarily delete some predicates, do some stuff, and then restore them" which I didn't like.

I agree that this commit does add more lines of code, which I don't like, but what I like less was the unofficial and haphazard methods we used.

When you say "this change", do you mean "adding index_predicates as an arg to this function"? We could probably adjust this. Or do you mean the larger-scale changes I mention above?

NickCrews:
@fgregg I fixed two of your requests, take a look at the final one that I didn't change because I wanted clarification.

dedupe/training.py
fgregg: those are all really good arguments. for me, the ideal thing would be for the `BlockLearner` classes to only have two public methods, the `__init__` and the `learn` methods. in particular, my initial instinct is that `cover` should not be called publicly.

i completely agree that we are violating that with how we are reaching into the objects with `self.blocker.block_learner.blocker.predicates.copy()`

i really like the rest of the changes in this PR. could we go back to the `self.blocker.block_learner.blocker.predicates.copy()` pattern for now, and open up separate issue for a design for that?

fgregg:
hi @NickCrews. sorry for the slow response!

I responded two to you on the `cover` change.

dedupe/training.py
fgregg: actually, i'm good with this!

fgregg:
@benchmark

dedupe/training.py
fgregg: i'm really struggling with this. i think this is better than the code that it is replacing, but it seems like this filtering should not be the responsibility of the BlockLearner class.

fgregg: hmm... what about this design.

```python
class BlockLearner(ABC):
    def learn(self, matches, recall, candidate_types='simple', selected_predicates=None):
        comparison_cover = self.comparison_cover
        if selected_predicates:
            candidate_preds = comparison_cover & selected_predicates
        else:
            candidate_preds = comparison_cover.keys()
   
        match_cover = self.cover(matches, candidate_preds)
        ...



    def cover(self, pairs, candidate_preds) -> Cover:
        predicate_cover = {}
        for predicate in candidate_preds:  # type: ignore
        ...
        
```

then this labeler code could just be

```python
    def learn_predicates(
        self, recall: float, selected_predicates
    ) -> tuple[Predicate, ...]:

        learned_preds: tuple[Predicate, ...]
        dupes = [pair for label, pair in zip(self.y, self.pairs) if label]
        learned_preds = self.blocker.block_learner.learn(
                dupes, recall=recall, candidate_types="random forest",
                selected_predicates
            )
```

and so then it would ultimately be the responsible for the `train` method in the api.py
 
to filter the predicates based and pass them on appropriately.

NickCrews: I'm trying to wrap my head around the general pattern of what we're doing, correct me if I'm wrong. Right now there are three sets of predicates:
1. all the available predicates from the data model
2. A subset of 1, all the predicates that the BlockLearners use as candidates.
3. A subset of 2, the final predicates chosen by the BlockLearners that we use to actually do the blocking in Dedupe.pairs().

Right now BlockLearners are handed 1, and it filters that down to 2 using whether or not predicates are index predicates or not, using the `index_predicates` flag. Then it does the step of filtering from 2 to 3.

I agree that this is bad, these concerns should be separated. The filtering from 1 to 2 should happen external to the BlockLearners. The block learners should just get handed set 2, and they select set 3 from that. That is the direction that https://github.com/dedupeio/dedupe/issues/1052 is trying to go in.

I'm not sure the 1->2 filtering should happen in train() though. Can't we do this at the very beginning, when the data model is first given to Dedupe? Or at least in training.DedupeBlockLearner.__init__, because there we call 

```
self.blocker = blocking.Fingerprinter(predicates)
self.blocker.index_all(data)
```

which makes us index ALL the predicates, even if then in subsequent calls to learn() or cover() we then filter out the index predicates and don't use them.

If you think this is at least an improvement, how about you merge it and then we can actually write PRs that go in this even better direction, and we have something concrete to talk about? I'm finding it hard without actually testing things out, but I don't want to have merge conflicts later so I want main to be stable.

NickCrews: But yes I think your suggestion is better than what I have currently.

NickCrews: This conversation is getting too big, should be it's own issue?

NickCrews: @fgregg in case you didn't see this

fgregg:
i'm going to go ahead and bring this in, and we can keep on working on improving how we interact with predicates in future PRs.

NickCrews:
Sweet, thanks for keeping momentum going @fgregg !

PR #1079, opened 2022-08-11
Extract predicate filtering from data model

NickCrews:
Part of the quest to remove the implementation details
of predicates out of DataModel and into the things that
actually care about them.

This slightly changes the behavior in the test because we don't
do any filtering either way, so we use ALL predicates from the
variable definitions

Inspired by https://github.com/dedupeio/dedupe/pull/1065#discussion_r943910477

@fgregg this is ready for review

### DIFF
--- a/dedupe/labeler.py
+++ b/dedupe/labeler.py
@@ -225,6 +225,20 @@ class BlockLearner(Learner):
         return sample_ids
 
 
+def _filter_canopy_predicates(
+    predicates: Iterable[Predicate], canopies: bool
+) -> set[Predicate]:
+    result = set()
+    for predicate in predicates:
+        if hasattr(predicate, "index"):
+            is_canopy = hasattr(predicate, "canopy")
+            if is_canopy == canopies:
+                result.add(predicate)
+        else:
+            result.add(predicate)
+    return result
+
+
 class DedupeBlockLearner(BlockLearner):
     def __init__(
         self,
@@ -239,7 +253,8 @@ class DedupeBlockLearner(BlockLearner):
 
         index_data = sample_records(data, 50000)
         sampled_records = sample_records(index_data, N_SAMPLED_RECORDS)
-        preds = self.data_model.predicates()
+        preds = self.data_model.predicates
+        preds = _filter_canopy_predicates(preds, canopies=True)
 
         self.block_learner = training.DedupeBlockLearner(
             preds, sampled_records, index_data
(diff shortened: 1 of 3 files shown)


dedupe/labeler.py
```diff
@@ -225,6 +225,20 @@ def _sample_indices(self, sample_size: int) -> Iterable[RecordIDPair]:
         return sample_ids
 
 
+def _filter_canopy_predicates(
+    predicates: Iterable[Predicate], canopies: bool
```
NickCrews, line 229: @fgregg this preserves the logic that was here before, but I wanted to check that this actually is the logic that we want. I was wondering if this was actually intending to filter out canopies? Because that's not what it currently does, if you pass canopies=True, then it only includes canopies, it filters out non-canopies.

eg perhaps this is intended to be:

```python
def _filter_canopy_predicates(
    predicates: Iterable[Predicate], canopies: bool
) -> set[Predicate]:
    if canopies:
        return set(predicates)
    result = set()
    for predicate in predicates:
        if hasattr(predicate, "index"):
            is_canopy = hasattr(predicate, "canopy")
            if is_canopy:
                result.add(predicate)
        else:
            result.add(predicate)
    return result
```

NickCrews:
@benchmark

NickCrews:
I have no idea why the tests are failing, since they weren't failing just a few commits ago and I don't see what could have changed. Looking at the logs I *think* all the dependencies stayed the same, but maybe I missed something. but dependencies shouldn't be related to this?

NickCrews:
I think editable install is failing due to https://github.com/pypa/setuptools/pull/3497?

fgregg:
i'm not sure that datamodel isn't the right owner of predicates, since it knows about the fields?

fgregg:
but i suppose the narrower question is whether we want datamodel to care about whether we want certain types of predicates.

fgregg:
can you rebase this so we can get tests going, @NickCrews ?

NickCrews:
> i'm not sure that datamodel isn't the right owner of predicates, since it knows about the fields?
> but i suppose the narrower question is whether we want datamodel to care about whether we want certain types of predicates.

I think I see what you're saying, and I think I agree. Yes, this PR is just slightly improving the situation.

NickCrews:
Can we get this merged since it's an improvement, and then come back later to fully extract predicates?

fgregg:
thanks for the ping! yes, let's do it.

PR #1081, opened 2022-08-11
Move project metadata to pyproject.toml

NickCrews:
It's encourage for project to define their metadata statically.
Makes static analyzers and other tooling work better.


NickCrews:
Superseded by #1082, where I lock setuptools==63, and then the tests pass.

PR #1082, opened 2022-08-11
Setuptools-63

NickCrews:
Testing if we're getting hit by
https://github.com/pypa/setuptools/issues/3504?


NickCrews:
Followup to #1081 with a fix for https://github.com/pypa/setuptools/issues/3504

fgregg:
@NickCrews, i like moving to setup.cfg, but i think we'll need to do #1085 to really address this.

NickCrews:
You want to wait for #1085 to land first or can we get this in and then come back later and fix it for real? Since it's just pinning a build requirement I don't think this should break anyone.

And are you fine with a pyproject.toml or do you want it in setup.cfg?

fgregg (changes requested):
i like this change. can you get the settings-file-persistence test working 

also, there's a couple settings in setup.cfg, it would be nice to transfer them to pyproject.toml and drop setup.cfg is that's possible

NickCrews:
> also, there's a couple settings in setup.cfg, it would be nice to transfer them to pyproject.toml and drop setup.cfg is that's possible

flake8 doesn't support pyproject.toml, so we're stuck with that as the only thing in the setup.cfg. Could use .flake8 if we wanted to relegate it to its own little world. If you want some juicy drama check out https://github.com/PyCQA/flake8/issues/234.

NickCrews:
I just rebased on main, and that solved the settings file persistence problem.

fgregg:
hmm yes, let’s do .flake8!

PR #1086, opened 2022-08-30
Auto cleanup memmap scores

NickCrews:
Users don't need to worry about cleaning up the memmpa file now


dedupe/api.py
```diff
@@ -175,7 +174,6 @@ def partition(
         clusters = self.cluster(pair_scores, threshold)
         clusters = self._add_singletons(data, clusters)
         clusters = list(clusters)
```
NickCrews, line 191: Now that we have auto-cleanup, perhaps we don't want to cast this to a list? I think then this entire method would be out-of-core, right?

Same for casting to list in join()

dedupe/core.py
```diff
@@ -176,9 +177,28 @@ def scoreDuplicates(
     else:
         scored_pairs = numpy.array([], dtype=dtype)
 
+    # See https://docs.python.org/3/library/weakref.html#comparing-finalizers-with-del-methods
+    scored_pairs.remove = weakref.finalize(scored_pairs, _cleanup_scores, scored_pairs)
+    scored_pairs.removed = property(_is_removed)
```
NickCrews, line 182: IDK if including the `removed` property is good, probably could live without it.

dedupe/core.py
```diff
@@ -176,9 +177,29 @@ def scoreDuplicates(
     else:
         scored_pairs = numpy.array([], dtype=dtype)
 
+    # Monkeypatch in these extra methods and attributes.
```
fgregg, line 180: this is really cool!

dedupe/api.py
fgregg: that's right. i think we we should put in a warning for a few releases though because some folks are likely depending on it being a list instead of a generator

fgregg:
wow! i've long wanted to do something like this but __del__ was always so finicky. a super nice approach.

can we get a test for this: something like

```
create_scored_pairs
get file location for memmap
del scored_pairs
check that there is no file at that location
```

fgregg (changes requested):
made a request for a test in the thread

dedupe/core.py
NickCrews: Someone could call pairs.remove() manually to do the finalization, and then it would be True. But you're correct, we as the library rely on garbage collection so we never take advantage of the .removed property. We could remove it, always easier to add later than to have to continue supporting it forever...

dedupe/api.py
NickCrews: do we want to do what Gazeteer.search() does with the `generator: bool` parameter?

```python
def score(.... generator: bool = False):
    ...
    if not generator:
        warn("this deprecated")
        return list(scores)
    else:
        yield from scores
```

dedupe/core.py
NickCrews: Yeah seriously! The first time I've used it too.

NickCrews:
Yes the test seems important, I can look into it

Issue #1088, opened 2022-09-01
Split up Datamodel into predicates, rename to Featurizer


NickCrews:
Branching off from thoughts at https://github.com/dedupeio/dedupe/pull/1079#issuecomment-1212621033

Datamodel currently has two responsibilities:
- holding candidate predicates
- calculating feature vectors between pairs

We should split this up. We could still generate these two things from variable definitions, but they should be stored separately. This would help for instance in labeler.py, when we are passing around a Datamodel willy nilly, when really what we care about is either the predicates for the BlockLearner, or the feature vectors for the RLRLearner. We should only pass around the banana, not the entire gorilla holding the banana.

By doing this, it would also allow us to better abstract the two separate steps of
- Generating feature vectors between pairs
- Using these feature vectors to calculate the 0-1 similarity score

For instance, I would like to be able to modify one of these steps, and it currently is a bit difficult to get in the middle and modify what's happening. To be precise, I want to add hard rules, such as "if both first name and address match, then mark similarity as 1" (I can add a custom variable for this that helps, but this doesn't guarantee that they get a score of 1).

fgregg:
i think that's a nice idea, nick! i would love to see that.

NickCrews:
If we remove the DataModel class, it's going to be difficult to remain backwards compatible with the settings file loading in api.py. eg the `self.data_model = pickle.load(settings_file)` I think we would need to keep some skeleton of the DataModel class around in the same location it is now, and add in some converters to this new architecture. Doable, but gross.

What do you think @fgregg, should I stay backwards compatible?

NickCrews:
Thinking about this more, I think it might make sense to combine the Featurizer and Classifier into one object, FeatureScorer, which is a concrete implementation of a Scorer protocol. Dedupe and the higher classes should only operate on the
Scorer protocol, and not know about how under the hood the pairs are getting featurized and classified:

```
class Scorer(Protocol):
    def score(self, pairs: Pairs) -> Scores:
        ...

class FeatureScorer(Scorer):
    def __init__(self, featurizer, classifier):
        self.featurizer = featurizer
        self.classifier = classifier

    def score(self, pairs):
        features = self.featurizer(pairs)
        return self.classifier(features)
```

By doing this, it allows someone (ie me :) ) to plug in their own scorer that operates in a different way. For instance, I have some hardcoded rules that I want to enforce ("if first names don't match, then give a score of 0") and this logic is impossible with the current featurize/classify pipeline we have now.

[...]

PR #1098, opened 2022-09-16
Reduce responsibilities of DataModel

NickCrews:


NickCrews:
This is the start of fixing https://github.com/dedupeio/dedupe/issues/1088

first, remove responsibilities of DataModel

second, simplify how DataModel works

third, prep for making it easier to make next changes backwards compatible by making the pickling shcema better versioned.

@fgregg there's some definitely good stuff in here (the first and second stuff), but I'm not sure if the third step is gonna be wasted effort. I thought I'd get a review from you at this point before I sink more time into this just o make sure I'm headed in the right direction. I can split this PR into three separete PRs too if you want. Thanks for the review!

dedupe/datamodel.py
```diff
@@ -183,30 +178,33 @@ def typify_variables(
...
-
-        if hasattr(variable_object, "higher_vars"):
-            all_variables.extend(variable_object.higher_vars)
-        else:
-            variable_object = cast(Variable, variable_object)
-            all_variables.append(variable_object)
+        assert isinstance(variable_object, Variable)
+        variables.append(variable_object)
 
+    only_custom = all(isinstance(v, (CustomType, InteractionType)) for v in variables)
```
fgregg, line 189: variable name should change. maybe `no_blocking_variables`

dedupe/datamodel.py
```diff
@@ -183,30 +178,33 @@ def typify_variables(
...
             "At least one of the variable types needs to be a type"
             "other than 'Custom'. 'Custom' types have no associated"
             "blocking rules"
         )
 
-    return primary_variables, all_variables
+    return variables
+
+
+def _expand_higher_variables(variables: Iterable[Variable]) -> list[Variable]:
```
fgregg, line 200: we could probably do away with this method if we had the base variable type have a higher_vars attribute set to an empty list or tuple

fgregg: no that won't quite work. it still feels like some of this should maybe be the responsible for the variable class.

fgregg (changes requested):
generally, these all look like good code cleanups. I think it would be good to split into three PRs!

dedupe/datamodel.py
NickCrews: It definitely feels gross, there needs to be some more official API. Trying not to deal with that here.

NickCrews:
Split into #1101 and #1102. I just dropped the 3 commits that adjusted the unpickling, I wasn't as confident those were going in the right direction.

dedupe/datamodel.py
NickCrews: Responded in followup pr #1102

PR #1102, opened 2022-09-23
Prep DataModel for removal

NickCrews:
This makes it easier to create Variable instances from variable definitions, because Variables are responsible for both predicates and distance functions. So, if we split DataModel into these two separate parts, we're going to need to do something like

```
# In eg Dedupe or RecordLink 
vars = typify_variables(var_defs)
predicates = [v.predicates() for v in vars]
featurizer = Featurizer(vars)
```

so it would be important that we can create our variable instances once, and then pass those around to the other parts.

### DIFF
--- a/dedupe/datamodel.py
+++ b/dedupe/datamodel.py
@@ -179,9 +179,9 @@ def typify_variables(
         try:
             variable_class = VARIABLE_CLASSES[variable_type]
         except KeyError:
+            valid = ", ".join(VARIABLE_CLASSES)
             raise KeyError(
-                "Field type %s not valid. Valid types include %s"
-                % (definition["type"], ", ".join(VARIABLE_CLASSES))
+                f"Variable type {variable_type} not valid. Valid types include {valid}"
             )
 
         variable_object = variable_class(definition)
--- a/dedupe/datamodel.py
+++ b/dedupe/datamodel.py
@@ -50,9 +50,6 @@ class DataModel(object):
 
         self._len = len(all_variables)
 
-    def __len__(self) -> int:
-        return self._len
-
     # Changing this from a property to just a normal attribute causes
     # pickling problems, because we are removing static methods from
     # their class context. This could be fixed by defining comparators
@@ -82,7 +79,7 @@ class DataModel(object):
     ) -> numpy.typing.NDArray[numpy.float_]:
         num_records = len(record_pairs)
 
-        distances = numpy.empty((num_records, len(self)), "f4")
+        distances = numpy.empty((num_records, self._len), "f4")
 
         for i, (record_1, record_2) in enumerate(record_pairs):
 
--- a/dedupe/datamodel.py
+++ b/dedupe/datamodel.py
@@ -217,13 +217,13 @@ def missing(variables: list[Variable]) -> list[MissingDataType]:
 def interactions(
     definitions: Iterable[VariableDefinition], primary_variables: list[FieldVariable]
 ) -> list[InteractionType]:
-    field_d = {field.name: field for field in primary_variables}
+    var_d = {var.name: var for var in primary_variables}
 
     interactions = []
     for definition in definitions:
         if definition["type"] == "Interaction":
             var = InteractionType(definition)
-            var.expandInteractions(field_d)
+            var.expandInteractions(var_d)
             interactions.extend(var.higher_vars)
     return interactions
 
--- a/dedupe/datamodel.py
+++ b/dedupe/datamodel.py
@@ -29,7 +29,7 @@ if TYPE_CHECKING:
     )
     from dedupe.predicates import Predicate
 
-VARIABLE_CLASSES = {k: v for k, v in FieldVariable.all_subclasses() if k}
+VARIABLE_CLASSES = {k: v for k, v in Variable.all_subclasses() if k}
 
 
 class DataModel(object):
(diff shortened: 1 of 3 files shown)


dedupe/datamodel.py
```diff
@@ -183,30 +178,35 @@ def typify_variables(
...
+        variables.append(variable_object)
 
-        if hasattr(variable_object, "higher_vars"):
-            all_variables.extend(variable_object.higher_vars)
-        else:
-            variable_object = cast(Variable, variable_object)
-            all_variables.append(variable_object)
-
-    if only_custom:
+    no_blocking_variables = all(
```
NickCrews, line 189: Continuing from https://github.com/dedupeio/dedupe/pull/1098#discussion_r975730820

I think it would be better if this was more duck-typed though, such as 
`no_blocking_variables = not any(len(v.predicates) for v in variables)`.

That would be a slight change though, since `Variable.predicates` is filled with an `ExistsPredicate` for any var def with the `has missing` flag. Perhaps this is OK? But I don't think so.

I think really there needs to be a more official API where variables declare the predicates they export.

dedupe/datamodel.py
```diff
@@ -39,11 +39,16 @@ def __init__(self, variable_definitions: Iterable[VariableDefinition]):
...
         if not variable_definitions:
             raise ValueError("The variable definitions cannot be empty")
-        self.primary_variables = typify_variables(variable_definitions)
-        all_variables = _expand_higher_variables(self.primary_variables)
-        self._derived_start = len(all_variables)
-
-        all_variables += interactions(variable_definitions, self.primary_variables)
+        variables = typify_variables(variable_definitions)
+        non_interactions: list[FieldVariable] = [
+            v for v in variables if not isinstance(v, InteractionType)  # type: ignore[misc]
```
NickCrews, line 41: This is another case where I think a Variable should have a more official API for if they export if they are an Interaction or not. Maybe should think of a better name for this too. Atomic vs derived? leaf vs nonleaf? base vs derived?

dedupe/datamodel.py
```diff
@@ -181,16 +181,18 @@ def typify_variables(
...
+        assert isinstance(variable_object, Variable)
 
         variables.append(variable_object)
 
-    only_custom = all(isinstance(v, CustomType) for v in variables)
-    if only_custom:
+    no_blocking_variables = all(
+        isinstance(v, (CustomType, InteractionType)) for v in variables
+    )
+    if no_blocking_variables:
```
NickCrews, line 192: Perhaps this check should happen in DataModel.__init__, and not in this utility function?

dedupe/datamodel.py
```diff
@@ -214,16 +216,12 @@ def missing(variables: list[Variable]) -> list[MissingDataType]:
     return missing_variables
 
 
-def interactions(
-    definitions: Iterable[VariableDefinition], primary_variables: list[FieldVariable]
-) -> list[InteractionType]:
-    var_d = {var.name: var for var in primary_variables}
-
+def _expanded_interactions(variables: list[Variable]) -> list[InteractionType]:
+    field_vars = {var.name: var for var in variables if isinstance(var, FieldVariable)}
```
NickCrews, line 221: Again, this should feels like it needs a more official API, or should be responsibility of the InteractionType

dedupe/datamodel.py
fgregg: the code that add the ExistsPredicate should only really be on FieldPredicates as it depends on there being a field.

https://github.com/dedupeio/dedupe/blob/main/dedupe/variables/base.py#L37

if we move that code, then we could do pretty much what you are saying.

```python
no_blocking_variables = not any(len(v.predicates) for v in variables)
```

we might also want to have a check like

```python
only_exists = all(isinstance(pred, ExistsPredicate) for v in variables for pred in v.predicates)
if only_exists:
    warning.warn('You are only using the Exists Predicate which is usually pretty bad')
```

dedupe/datamodel.py
fgregg: it seems of a piece with the "Incorrect variable specification: variable" error?

dedupe/datamodel.py
```diff
@@ -183,30 +178,35 @@ def typify_variables(
...
+
+
+def _expand_higher_variables(variables: Iterable[Variable]) -> list[Variable]:
+    result: list[Variable] = []
+    for variable in variables:
+        if hasattr(variable, "higher_vars"):
+            result.extend(variable.higher_vars)
+        else:
+            result.append(variable)
+    return result
```
fgregg, line 209: what if we added an __iter__ method to Variable so that we could just do something like.

```
features = []
for variable in variables:
   for feature in variable:
       features.append(feature)
```

dedupe/datamodel.py
fgregg: i can think of a few designs here, but this code is actually doing two things for us, both of which are kind of important.

1. isolating the interaction features for us, so we can put them in the right place in the sequence of features. 
2. actually resolving the feature indices. 

seems like we kind of need to do both.

dedupe/variables/base.py
```diff
@@ -58,21 +58,16 @@ def all_subclasses(
 
 
 class DerivedType(Variable):
-    type = "Derived"
```
fgregg, line 61: why did you remove this?

dedupe/variables/base.py
```diff
@@ -58,21 +58,16 @@ def all_subclasses(
...
-
     def __init__(self, definition: VariableDefinition):
         self.name = "(%s: %s)" % (str(definition["name"]), str(definition["type"]))
         super(DerivedType, self).__init__(definition)
 
 
 class MissingDataType(Variable):
-    type = "MissingData"
```
fgregg, line 69: why did you remove this?

fgregg (changes requested):

fgregg:
thanks for working through this @NickCrews. this is some of the first code i wrote for project and there's a lot i didn't know!

dedupe/datamodel.py
NickCrews: I think this is a good idea in general, but I would want to name it to differentiate between predicates and features, since Variables supply both. Perhaps `Variable.features`?

dedupe/variables/base.py
NickCrews: I might be wrong but my understanding is:

This would only be used if someone passes in a variable definition like `{"type": "Derived"}`, and then we would create an instance of DerivedType, but that doesn't make sense because DerivedType is an abstract base class and shouldn't ever be created directly.

PR #1107, opened 2022-09-29
udpate readme for cloning repo

f-hafner:
Small thing, but it may save some others some time. I think 

```bash
git clone git://github.com/dedupeio/dedupe.git
```
should be 

```bash
git clone https://github.com/dedupeio/dedupe.git
```

At least I got a time out error because of this I believe: https://github.blog/2021-09-01-improving-git-protocol-security-github/

Thought was fastest to directly fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Once you have virtualenvwrapper set up,
 
 ```bash
 mkvirtualenv dedupe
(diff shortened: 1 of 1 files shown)


fgregg:
thanks!

PR #1114, opened 2022-11-16
Unmap file to prevent PermissionError when deleting temp file on Win

PaulM5406:
On Windows, a PermissionError is raised when trying to removing the temporary file:
`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: ...`

Unmap the temp file before deleting it prevents this error.


fgregg:
thanks!!

PR #1115, opened 2022-11-18
Fix incompatible change for random.sample() in python 3.11

lmores:
In Python 3.11 the `population` argument for `random.sample(population)` must be a sequence, automatic conversion is no longer supported, see the [documentation](https://docs.python.org/3/library/random.html#random.sample).

### DIFF
--- a/dedupe/labeler.py
+++ b/dedupe/labeler.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import random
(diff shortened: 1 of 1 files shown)


fgregg:
thanks so much @lmores, i took your fix in https://github.com/dedupeio/dedupe/commit/d1550dce3449842859e5cd84f282677b819de5e8

Issue #1085, opened 2022-08-17
transition to plugins for dedupe variables.


fgregg:
newer versions of pip/setuptools are now confused by the [old-style namespace packages](https://github.com/pypa/setuptools/issues/3537#issuecomment-1218204087).

i previously tried to migrate to the implicit style namespace packages, but the problem with that i don't think there's a way to do that this not backwards incompatible. 

specifically, i don't see how you could use the implicit style namespaces and have something like `from dedupe import Dedupe` work. 

you could do `from dedupe.api import Dedupe` but that's not what we have guaranteed.

so, then maybe we have to move to the [src-layout](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#src-layout)

NickCrews:
I don't see a problem with just using a src layout?

NickCrews:
Also, what was "backwards incompatible" with it? Like any custom variables that got registered with dedupe by being implicitly part of the `dedupe` package wouldn't get registered anymore?

I would say for that I don't currently like how that registration is implicit via subclassing. I think it would be much better to have explicit @dedupe.register_variable decorator that people could put on their own classes, or even use a plugin system like [pluggy](https://pluggy.readthedocs.io/en/stable/)

fgregg:
with implicit namespaces you can't have an __init__.py in the top directory as far as i understand.

so everything that's under dedupe right now would have to go into a subdirectory like "core" whatever.

the reasons why we use subclassing is that was the main way to do it when we first wrote the code way back. 

there's not really any path forward that's not going to be a breaking change for existing plugins, so i think we should feel free to do what we want. i'm very open to more modern approaches here.

fgregg:
@NickCrews, do you have an example of a use of pluggy that would be a good fit for dedupe variables? most of what i'm finding is about hooks, which doesn't seem quite right.

NickCrews:
I don't. And thinking about it more, pluggy seems a bit overkill.

What about we do what pandas does with their plotting backend? https://plotly.com/python/pandas-backend/

Eg they say `pandas.options.plotting.backend = "plotly"` and pandas goes and tries to import a package named "plotly" and thn use a function called "plot" in that package.

So if you specify the type of a variable as 
"mylib.foo:ZipCodeVariable" then dedupe imports mylib.foo and looks for an attribute named ZipCodeVariable. Using a colon as a separator as is used as a convention elsewhere in python eg setuptools entry points.

[...]

PR #1122, opened 2022-12-11
Reference plugin variables using "module:class" strings

NickCrews:
Will close https://github.com/dedupeio/dedupe/issues/1085

This probably could do with some more polishing, eg removing `type` from every Variable class, and swapping `if variable_type == "Interaction"` to parsing the variable type, then checking`if isinstance(variable_type, dedupe.variables.InteractionType)`

### DIFF
--- a/dedupe/datamodel.py
+++ b/dedupe/datamodel.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import copyreg
 import pkgutil
 import types
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Type, cast
 
 import numpy
 
@@ -142,6 +142,16 @@ class DataModel(object):
         self.__dict__ = d
 
 
+def _get_variable_class(variable_type: str) -> Type[Variable]:
+    try:
+        return VARIABLE_CLASSES[variable_type]
+    except KeyError:
+        raise KeyError(
+            "Field type %s not valid. Valid types include %s"
+            % (variable_type, ", ".join(VARIABLE_CLASSES))
+        )
+
+
 def typify_variables(
     variable_definitions: Iterable[VariableDefinition],
 ) -> tuple[list[FieldVariable], list[Variable]]:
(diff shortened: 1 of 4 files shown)


fgregg:
i like the simplicity of this approach, but it will be a breaking change for data models that include datetime. in this code this is only working because of the surviving name spacing bits we are trying to get away from.

it will also be a be a breaking change for data models that include the name, address, and fuzzy category types that are not bundled with dedupe but are recommended in the docs

NickCrews:
> i like the simplicity of this approach, but it will be a breaking change for data models that include datetime. in this code this is only working because of the surviving name spacing bits we are trying to get away from.

Hmmm, you're right. Two possible solutions I see:

1. Merge the datetime package into this library. It's already listed as a dep of dedupe, and I assume that it isn't used by some 3rd package. I don't understand why datetime is given this special treatment but name, address, and fuzzycategory are not?
2. Update the datetime package so that it exports itself in a more normal way. Then here in `variables/__init__.py` we can say `from dedupe_variable_datetime import DateTimeType as DateTime`

> it will also be a breaking change for data models that include the name, address, and fuzzy category types that are not bundled with dedupe but are recommended in the docs

Hmm, that's a good point. I was just thinking of breaking changes to people using their own custom plugins. I think the best option here is temporarily have a hardcoded mapping for these edge cases (eg "FuzzyCategorical"->"dedupe_variable_fuzzycategory:FuzzyCategorical"), and if you use them then you get a deprecation warning. We can remove them later, after users get a chance to switch. I think eventually they should follow the same rules as other 3rd party plugin variables, since they are external and need to be pip installed.

> while this pattern is nicer for devs, i don’t like that if a library’s file structure changes, that can break a downstream users code. the file structure should not have to be part of an external API (though i grant that it often is)

That's a good point. I think that is the responsibility of the plugin authors and is easily avoided. e.g. in a `zipcodelib` package, in the root `__init__.py`, they should `from zipcodelib.foo.bar import ZipCodeVariable`, so that users can just use "zipcodelib:ZipVariable". Then the plugin author is free to move ZipCodeVariable from zipcodelib/foo/bar.py to zipcodelib/baz.py, adjust the import in the root `__init__.py`, and everything still works.

fgregg:
i think my main reason for preferring this is consistency of interface. i don't like that there are some variables that we refer to with a short alias and others that we use a path in the datamodel.

we could have a deprecation path where the paths were *the* way to reference variables, but honestly i think that a lot of new people would be confused by that.

if we were going to go down that road, i might even prefer a datamodel that went from:

```python
data_model = [{field: "name", type: "String"},
              {field: "address", type: "String"}]
```

to 

```
from dedupe.variables import StringVariable
data_model = [StringVariable(field="name"), StringVariable(field="address")]
```

which would avoid the plugin question altogether.

i kind of like that solution, but it seems like a lot of disruption without a huge benefit to users.

fgregg:
closed by #1193

PR #1138, opened 2023-01-17
purify id types

### LINKED ISSUE #1136: Improve typing of Data
[`Data`](https://github.com/dedupeio/dedupe/blob/f0503e09a57e9d655cb48cd7b162b08d0af671af/dedupe/_typing.py#L53) is an important type in the library, but it's typing definition is a bit wrong.

https://github.com/dedupeio/dedupe/blob/f0503e09a57e9d655cb48cd7b162b08d0af671af/dedupe/_typing.py#L47-L53

it ought to be 

```python
Data = Union[Mapping[int, RecordDict], Mapping[str, RecordDict]]
```

but simply making that change will break many things in typing.

fgregg:
When I was using dedupe in another project, i noticed that we had, in many places, said we expected record ids to be a mixture of integers or strings. In reality, we need the record ids to either be purely integers or purely strings. This fixes that typing.

It's not quite done, because it is possible for the two datasets in RecordLinkage and Gazetteer code to have different id types, and i haven't handled that possibility yet.

closes #1136

### DIFF
--- a/dedupe/_typing.py
+++ b/dedupe/_typing.py
@@ -33,38 +33,52 @@ if TYPE_CHECKING:
 RecordDict = Mapping[str, Any]
 RecordID = Union[int, str]
 RecordIDDType = Union[Type[int], Tuple[Type[str], Literal[256]]]
-RecordIDPair = Tuple[RecordID, RecordID]
-Record = Tuple[RecordID, RecordDict]
-RecordPair = Tuple[Record, Record]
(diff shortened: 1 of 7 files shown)


fgregg:
> It's not quite done, because it is possible for the two datasets in RecordLinkage and Gazetteer code to have different id types, and i haven't handled that possibility yet.

actually, because of the way that the distance functions interact with numpy, the id types do need to be homogenous.

PR #1141, opened 2023-01-21
Change IDF formula

lmores:
Implement the change being discussed in #1140.

Also directly import `numpy.log` to avoid attribute access inside for loop.


lmores:
Experimental verification required.

fgregg:
@benchmark

fgregg:
it looks like it did maybe improve precision and recall for the restaurant dataset. why don't you re-open this and i can take a look at some other datasets.

PR #1145, opened 2023-01-22
Improve cpredicates.pyx

lmores:
Changes:
  * Avoid nested for loop in ngrams() function inside cpredicates.pyx.
  * Fix docstrings to match actual implementation.
  * Add tests.

Not sure how to check runtime improvement, using `python benchmarks/benchmarks/canonical.py` execution time is `11,xxx` seconds both before and after the change.

@fgregg: I am likely to open many more PR like this. Please tell me if you are fine with them. Of course, if I plan to submit bigger changes I will open a thread to discuss them before actually implementing them.

### DIFF
--- a/dedupe/cpredicates.pyx
+++ b/dedupe/cpredicates.pyx
@@ -39,7 +39,6 @@ cpdef tuple initials(basestring field, int n):
     return (ufield[:n], )
 
 
-
 cdef unicode _ustring(basestring s):
     if type(s) is unicode:
         # fast path for most common case(s)
--- a/dedupe/cpredicates.pyx
+++ b/dedupe/cpredicates.pyx
@@ -1,37 +1,33 @@
 # cython: c_string_type=unicode, c_string_encoding=utf8, infertypes=True, language_level=3
 
 cpdef list ngrams(basestring field, int n):
-    """ngrams returns all unique, contiguous sequences of n characters
+    """ngrams returns all contiguous sequences of n characters
     of a given field.
-        
-    :param field: the string to be 
+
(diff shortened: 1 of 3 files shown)


fgregg:
the uniqueness guarantee is required

lmores:
> the uniqueness guarantee is required

But was not enforced, right?

fgregg:
it was enforced by the set object

fgregg:
hmm that’s true

lmores:
Sorry, but I don't understand. At the moment I see not set() object inside cpredicates.pyx, in particular [ngrams](https://github.com/dedupeio/dedupe/blob/baa6071dc563a9440d897a8d2d04ba5cf8867255/dedupe/cpredicates.pyx#L17) is a list.
I can make it a set if necessary.

fgregg:
you are right that the current code doesn’t enforce uniqueness, i’ll have to check where that is enforced

fgregg:
everywhere we call this in predicates.py, we call set on it. 

it's a bit silly to do this. let's have cpredicates fill out a set and then not have those set calls in predicates.py.

lmores:
Actually there is one exception:

```python
class TfidfNGramPredicate(IndexPredicate):
    def preprocess(self, doc: str) -> Sequence[str]:
        return tuple(sorted(ngrams(" ".join(strip_punc(doc).split()), 2)))
```

But we probably want the ngrams to be unique also here?

lmores:
How about the `unique_ngrams` function I added in the last commit?

fgregg:
looks good!

PR #1147, opened 2023-02-10
Change predicate function signatures

lmores:
This PR implements the changes discussed in #1146 and some consequences thereof.

### Changes contained in this PR

  - Predicate functions have been moved to a separate file named `predicate_functions.py`
  - New tests for nearly all predicate functions have been written in `test_predicate_functions.py`
  - Old tests for predicate functions inside `test_predicates.py` have been moved to `test_predicate_functions.py`
  - The implementation of `oneGramFingerprint()` and `twoGramFingerprint()` has changed (should be better now), without affecting the output. Inside `test_predicate_functions.py` there are a few assertions that compare the ouput of the new and the old implementations.
  - The signature of predicate functions has become:
    ```python
    PredicateOutput = Union[set[str], tuple[str], tuple[()]]
    PredicateFunction = Callable[[Any], PredicateOutput]
    ```
    We agreed that each predicate function should return a `set` since we want unique items and we do not care about the order.
    However, as explained at the beginning of `predicate_functions.py`, the starting capacity of sets in CPython implementation is 4.
    ```python
    >>> sys.getsizeof({"a"})
    216
    >>> sys.getsizeof({"a","b","c","d"})
    216
    >>> sys.getsizeof({"a","b","c","d","e"})
    472
    ```
    Since many predicate functions always return collections with at most one item, always using sets would require much more memory than actually needed.
    To avoid wasting memory we return tuples with just one (or zero) items whenever the output of a predicate function never contains more than one item and we return actual sets otherwise.
  - Optimized loops in `training.py` and `labeler.py`.
    In both of them we had to check whether the output of two predicates had non empty intersection.
    The old way to do so was to build `set(pred_1_output) & set(pred_2_output)` and check whether it is empty or not.
    The new approach is to check whether they share at least one item, without computing the whole intersection.
    Since `pred_1_output` and `pred_2_output` are sets or tuples with at most one item, it is "fast" to loop throught the smallest one and check wheter at least one of its items is contained in the other.
  - In `predicate_functions.py` changed
    ```python
    two_start_words = re.compile(r"^([\w']+\s+[\w']+)").match
    ```
    to
    ```python
    two_start_words = re.compile(r"^([\w']+\W+[\w']+)").match
    ```
    in this way strings like `"go-away"` are considered to be made of two tokens, instead of just one (I think this is the expected behaviour).

### TO DO?

  - To further improve memory usage we could always return the same instance of a frozenset whenever we must return an empty collection.
    However this would rely on the assumption that the output of a predicate function is never modified, just read.
    Is it true???
  - Implementing all predicate function in Cython would significantly improve the performance? (in another PR)
  - Implement the trick used in `training.py` and `labeler.py` in Cython? (in another PR)

@fgregg: as you see I took the liberty to apply some more changes other than those we discussed in #1146, let my know if you agree on them or if I have to revert something.


lmores:
Is there a way to run all CI tests locally (including all different python version) to avoid to push and find out problem on GitHub? (sorry for the noob question)

lmores:
Sorry, for the mess of force pushes, seems that I fixed all problems except those raised by mypy and black: on my machine they run fine without errors, so I can not reproduce (and fix) them.

dedupe/predicate_functions.py
```diff
@@ -0,0 +1,268 @@
+import re
+from itertools import chain
+from math import copysign, floor, log10
+from typing import Any, Sequence, Set, Tuple, Union
+
+from doublemetaphone import doublemetaphone
+
+from dedupe.cpredicates import initials, ngrams, unique_ngrams  # noqa: F401
+
+# Each predicate function returns a collection of strings. Since order
```
fgregg, line 10: this will need to be revised.

fgregg (commented):
thanks for the pull request! I think it's going in a good direction. Let's just have a predicate always return a frozenset.

dedupe/training.py
```diff
@@ -156,17 +157,32 @@ def cover(self, pairs: TrainingExamples, index_predicates: bool = True) -> Cover
...
                 pred for pred in self.blocker.predicates if not hasattr(pred, "index")
             ]
+
+        _non_empty_intersection = BlockLearner._non_empty_intersection
         for predicate in predicates:
             coverage = frozenset(
                 i
                 for i, (record_1, record_2) in enumerate(pairs)
-                if (set(predicate(record_1)) & set(predicate(record_2, target=True)))
+                if _non_empty_intersection(
```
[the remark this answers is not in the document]
lmores: Done

dedupe/labeler.py
```diff
@@ -141,19 +142,29 @@ def learn_predicates(
...
+        _non_empty_intersection = BlockLearner._non_empty_intersection
         for record_1, record_2 in pairs:
-
             for predicate in self.current_predicates:
                 keys = predicate(record_2, target=True)
-                if keys:
-                    if set(predicate(record_1)) & set(keys):
-                        labels.append(1)
-                        break
+                if keys and _non_empty_intersection(predicate(record_1), keys):
```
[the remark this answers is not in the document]
lmores: Done

dedupe/_typing.py
```diff
@@ -79,7 +80,8 @@
 ComparisonCoverInt = Dict["Predicate", FrozenSet[Tuple[int, int]]]
 ComparisonCoverStr = Dict["Predicate", FrozenSet[Tuple[str, str]]]
 ComparisonCover = Union[ComparisonCoverInt, ComparisonCoverStr]
-PredicateFunction = Callable[[Any], Iterable[str]]
+PredicateOutput = Union[Set[str], Tuple[str], Tuple[()]]
```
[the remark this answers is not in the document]
lmores: Done.

dedupe/predicate_functions.py
lmores: I completely removed this comment as now we always return frozensets.

dedupe/labeler.py
```diff
@@ -142,13 +142,11 @@ def learn_predicates(
...
         labels: Labels = []
         for record_1, record_2 in pairs:
-
             for predicate in self.current_predicates:
                 keys = predicate(record_2, target=True)
-                if keys:
-                    if set(predicate(record_1)) & set(keys):
-                        labels.append(1)
-                        break
+                if not keys.isdisjoint(predicate(record_1)):
```
fgregg, line 147: this looks great. 

we need to wrap this in a try except block, because third-party variables may not be returning a frozenset.

something like
```python
            try:
                if not keys.isdisjoint(predicate(record_1)):
                   labels.append(1)
                   break
             except AttributeError:
                warn('the predicate {} is not returning a frozen set, this will soon be required behaviour'.format(predicate.__name__), DeprecationWarning)
                if not set(keys).isdisjoint(predicate(record_1)):
                   labels.append(1)
                   break
```

fgregg: we'll need to do this everywhere we expect predicates to return a set.

fgregg (changes requested):
looking good. couple more changes.

dedupe/predicate_functions.py
```diff
@@ -50,44 +17,44 @@
...
 
-def firstTokenPredicate(field: str) -> Union[Tuple[str], Tuple[()]]:
+def firstTokenPredicate(field: str) -> FrozenSet[str]:
     first_token = start_word(field)
     if first_token:
-        return first_token.groups()  # type: ignore
+        return frozenset(first_token.groups())
     else:
-        return ()
+        return EMPTY_FROZENSET
```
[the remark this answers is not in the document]
lmores: Done. I am curious why you prefer to create a new frozenset each time. Is it just simplicity over efficiency?

dedupe/labeler.py
lmores: Done everywhere.

fgregg:
just rebase this onto main so we can make sure the linting passes and we’ll be good to go

dedupe/predicate_functions.py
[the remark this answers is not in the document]
fgregg: simplicity

fgregg:
@benchmark

dedupe/predicates.py
```diff
@@ -116,13 +112,13 @@ def __init__(self, field: str):
         self.field = field
 
     @staticmethod
-    def func(column: Any) -> tuple[Literal["0", "1"]]:
+    def func(column: Any) -> set[Literal["0", "1"]]:
```
fgregg, line 115: wrong signature. should be frozenset

dedupe/predicates.py
```diff
@@ -116,13 +112,13 @@ def __init__(self, field: str):
...
         if column:
-            return ("1",)
+            return {"1"}
         else:
-            return ("0",)
+            return {"0"}
 
-    def __call__(self, record: RecordDict, **kwargs) -> tuple[Literal["0", "1"]]:
+    def __call__(self, record: RecordDict, **kwargs) -> set[Literal["0", "1"]]:  # type: ignore
```
fgregg, line 121: wrong signature. should be frozen set

fgregg (commented):
couple places where you missed the signature

dedupe/predicates.py
```diff
@@ -71,7 +67,7 @@ def __len__(self) -> int:
         return 1
 
     @abc.abstractmethod
-    def __call__(self, record: RecordDict, **kwargs) -> Iterable[str]:
+    def __call__(self, record: RecordDict, **kwargs) -> FrozenSet:
```
fgregg, line 70: the should be Frozenset[str] here and everywhere else

fgregg (commented):
another typing change

dedupe/predicates.py
lmores: Done

dedupe/predicates.py
lmores: Done

dedupe/predicates.py
lmores: Done

fgregg:
i'm a bit concerned by the slow down in what is supposed to be a performance improving PR. let's run @benchmark again

lmores:
Mmm, with frozensets in place I do no more expect any performance improvement as nothing in the long comment that I placed at the begging of predicate_functions.py do apply anymore.

fgregg:
i thought your argument was that not casting to sets in the labeler routines were where we would see performance gains?

fgregg:
@benchmark

lmores:
Yes, you are right, there was also that argument... could it be that frozensets are slightly less performant than sets in CPython? (don't know, just wondering)

fgregg:
okay that slow down was just line noise. i think this a nice improvement and standardization even if it's not a clear performance win.

PR #1152, opened 2023-03-17
Add support for record id of bytes type

lmores:
Add support for id columns of type `bytes`.

Reason: when the primary key of a row is the output of a hashing function, the most natural data type used for the corresponding database column is simply raw bytes (e.g., the BYTEA data type in postgreSQL).

I have already successfully used this patched version of dedupe to deduplicate rows from a postgresSQL table having the primary column of type BYTEA.

Let me know if:
  1. you are interested in adding support for id columns with `bytes` type
  2. you prefer to keep this change under the hood (as it is now) or make it explicit and update all the typing definitions in `_typing.py`


lmores:
@fgregg: any thoughts on this?

PR #1193, opened 2024-06-19
more direct set up data model, tests passing

### LINKED ISSUE #1085: transition to plugins for dedupe variables.

fgregg:
set up data models like

```python
[
  dedupe.variables.String("name"),
  dedupe.variables.Exact("address")
 ]
 ```

instead of 

```
[
  {"field", "name", "type": "String"},
  {"field", "address", "type": "String"},
 ]
```

supersedes #1122, #1121. will close #1085

Todo:
- [x] get tests passing
- [x] update documentation
- [x] update changelog
- [x] update external dedupe variables
- [x] raise helpful message if trying to initialize with a list of dicts
- [x] update example code

This is is going to be a breaking change, so do we want to make bigger changes like splitting datamodel into feature creators and blocking rules, a la #1102?


fgregg:
I still think @NickCrews's ideas about breaking apart the DataModel object is good, but i think that we should be able to do that in a way that does not break backwards compatibility, so I think we don't need to try to bring it in now.

NickCrews:
@fgregg dedupe returns from the dead! 😉 Great to see some love happening here, thanks for your work!

PR #1199, opened 2024-07-01
Add __len__ method to ExistType

ArVar:
Adds __len__ method to the `ExistType` like in `CategoricalType`. Otherwise the  ExistType will use the `__len__` method of `Variable`:
```
    def __len__(self) -> int:
        return 1
```

and higher_vars won't be expanded correctly in `datamodel.py`:

```
        for variable in self.field_variables:
            if len(variable) == 1:
                columns.append(variable)
            elif len(variable) > 1:
                assert hasattr(variable, "higher_vars")
                columns.extend(variable.higher_vars)
```

### DIFF
--- a/dedupe/variables/exists.py
+++ b/dedupe/variables/exists.py
@@ -29,7 +29,7 @@ class ExistsType(FieldType):
             return self.cat_comparator(0, 1)
         else:
             return self.cat_comparator(0, 0)
-        
+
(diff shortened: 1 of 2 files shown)


ArVar:
Hi @fgregg, are the codecov rules a bit too strict?
I haven't seen any test for __len__ in the repo.

